PR #6, opened 2020-01-22
Add target namespaces in operator group

horis233:
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/35225


deploy/olm-catalog/common-service-operator/0.0.3/common-service-operator.v0.0.3.clusterserviceversion.yaml
```diff
@@ -182,6 +182,8 @@ spec:
         - apiGroups:
           - operator.ibm.com
           - operators.coreos.com
+          - etcd.database.coreos.com
```
gyliu513: why add those two apigroups?

horis233: These two apigroups are just for installing the example operators etcd and Jenkins and their cr.  These two apigroups will be removed after integrate with common services operators.

chenzhiwei:
Removed the binary from this repo, all the commits are affected.

The size is significantly reduced:

```
128M	old
748K	new
```

You need to clone this repo again and update patch, then submit new PR.

PR #15, opened 2020-01-26
Push csv to quay.io

DanielXLee:
issue: https://github.com/IBM/common-service-operator/issues/9


horis233:
@DanielXLee 
We need to run `make check` `make test` `make build` to self check to make it pass the CICD check.

PR #16, opened 2020-01-27
Update OperatorGroup and Subscriptions when exist

horis233:
@DanielXLee 
I think we need to update the operator group and subscription when they exist to support the upgrade scenario and if users want to operator group and subscription after install.


horis233:
Will make a new one

DanielXLee:
@horis233 operator group generated by set controller, there is no interface for user change it, I don't think we need to consider this case.
For subscription updates, only the `channel` changed, do you think?

horis233:
@DanielXLee 
For implement common services dependency:   https://github.ibm.com/IBMPrivateCloud/roadmap/issues/35225#issuecomment-17273029

I add a `targetNamespaces` field into the metaOperator. I think I can use the same logic as the subscription. If the  `targetNamespaces`  is changed, the controller will update the operator group.

PR #20, opened 2020-01-30
Update operator group when targetnamespaces change

horis233:


DanielXLee:
@horis233 I don't think we need to consider update operator group. For the update case, @chenzhiwei @gyliu513 do we allow users to change the operator namespace?

horis233:
@DanielXLee 
Because we will use the target namespace in the operator group to manage the service dependency, I consider the case the dependency relationship is changed or users input a wrong target namespace. Then, I think we need to provide a chance to update the operator group.

PR #27, opened 2020-02-04
Add an example for installing common service operator on k8s

### LINKED ISSUE #25: Add document for how to use and test with vanila Kubernetes
We can provide an example with https://github.com/kubernetes-sigs/kind

FYI @horis233

horis233:
Fixed #25


docs/install_on_kubernetes.md
```diff
@@ -0,0 +1,204 @@
+# Install the common service operator on vanila Kubernetes
+
+In this document, we will show you how to deploy and use the meta operator on the vanila Kubernetes.
```
gyliu513: one minor comment, please use `doctoc` generate a ToC. We can fix this in another PR.

PR #51, opened 2020-02-07
Pass scorecard test

### LINKED ISSUE #45: pass operator-sdk scorecard test
https://github.com/operator-framework/operator-sdk/blob/master/doc/test-framework/scorecard.md
Test command:
```
operator-sdk scorecard -o text --selector='test in (checkspectest,checkstatustest,writingintocrshaseffecttest,bundlevalidationtest,crdshavevalidationtest,crdshaveresourcestest,specdescriptorstest,statusdescriptorstest)'
```

horis233:
fixed #45


deploy/operator.yaml
```diff
@@ -12,11 +12,11 @@ spec:
       labels:
         name: meta-operator
     spec:
-      serviceAccount: meta-operator
+      serviceAccountName: meta-operator
```
gyliu513: Do we need to change code after the spec change? @DanielXLee @horis233

horis233: Screen Shot 2020-02-09 at 9 21 56 PM
`serviceAccount` is depreciated, so I change it to `serviceAccountName`.
I don't think we need to change the operator code for it.

PR #69, opened 2020-02-12
Onboard cert-manager-operator into meta-operator

Crystal-Chun:
A few questions:
- Would the `channel` be `stable`? 
- What would `sourceName` and `sourceNamespace` be? I'm not sure what `CatalogSource` is


deploy/crds/operator.ibm.com_v1alpha1_metaoperatorcatalog_cr.yaml
```diff
@@ -20,3 +20,12 @@ spec:
     sourceNamespace: openshift-marketplace
     targetNamespaces:
       - etcd-operator
+  - name: ibm-cert-manager-operator
+    namespace: ibm-cs-certmanager
+    channel: <unsure>
+    packageName: ibm-cert-manager-operator
+    sourceName: <unsure>
+    sourceNamespace: <unsure>
+    targetNamespaces:
```
gyliu513: @horis233 @chenzhiwei I think we will not need `targetNamespace` as we are planning to listen to all namespaces.

@Crystal-Chun we also need to add a `description` field here to clarify some info for your operator, like dependency etc.

Crystal-Chun: Thanks for the review @gyliu513! I've removed `targetNamespaces` and added a `description` field.

Two questions about `description`:
1. What is the format of it? 
2. What types of dependencies do we need to add? Cert-manager doesn't depend on other common services.

horis233:
Channel should be alpha https://github.com/IBM/ibm-cert-manager-operator/blob/master/deploy/olm-catalog/ibm-cert-manager-operator/ibm-cert-manager-operator.package.yaml#L3

`sourceName` and `sourceNamespace` are used for getting catalog source. After you push your operator to https://quay.io/application/. You can set them to `opencloud-operators` and `openshift-marketplace`

deploy/crds/operator.ibm.com_v1alpha1_metaoperatorcatalog_cr.yaml
gyliu513: > What is the format of it?

A string is good enough

> What types of dependencies do we need to add? Cert-manager doesn't depend on other common services.

OK, then adding some simple description for your service is good enough.

Crystal-Chun:
Thanks @horis233! I've changed the channel to reflect that. 

Do I push the image manually to https://quay.io/application/ or is that done via a robot account?

deploy/crds/operator.ibm.com_v1alpha1_metaoperatorcatalog_cr.yaml
Crystal-Chun: Ok great, I'll remove the Kubernetes part of it. Thank you!

horis233:
I think you need to manually do that. please follow https://github.com/operator-framework/community-operators/blob/master/docs/testing-operators.md#push-to-quayio

Crystal-Chun:
I've got everything ready to push to quay.io/applications, but I no longer have push access. There's a script Steve created for pushing it through the prow builds, though! I'm going to integrate that into my repository.

deploy/crds/operator.ibm.com_v1alpha1_metaoperatorcatalog_cr.yaml
```diff
@@ -20,3 +20,10 @@ spec:
     sourceNamespace: openshift-marketplace
     targetNamespaces:
       - etcd-operator
+  - name: ibm-cert-manager-operator
+    namespace: ibm-cs-certmanager
```
chenzhiwei: can you set it to `ibm-cert-manager-operator`?

deploy/crds/operator.ibm.com_v1alpha1_metaoperatorconfig_cr.yaml
```diff
@@ -13,3 +13,9 @@ spec:
     spec:
       etcdCluster:
         size: 1
+  - name: ibm-cert-manager-operator
+    spec:
+      enableWebhook: true
+      imageRegistry: quay.io/opencloudio
+      pullSecret:
```
chenzhiwei: can you remove this? instead you can put this into cert-manager operator CSV file.

Later when you want to change something, you don't need to notify meta-operator.

Crystal-Chun: Do I put `certManager: {}` instead of all of this instead?

deploy/crds/operator.ibm.com_v1alpha1_metaoperatorcatalog_cr.yaml
Crystal-Chun: Made and committed the change

horis233:
@Crystal-Chun Could you add cert-manager into the `alm-example` in the meta-operator csv?

https://github.com/IBM/meta-operator/blob/master/deploy/olm-catalog/meta-operator/0.0.1/meta-operator.v0.0.1.clusterserviceversion.yaml#L5-L101

Crystal-Chun:
Thanks for reviewing! @horis233 I've added the cert-manager to the `alm-example`.

PR #73, opened 2020-02-13
update crs to onboard metering

sgrube:


deploy/crds/operator.ibm.com_v1alpha1_metaoperatorconfig_cr.yaml
```diff
@@ -4,6 +4,16 @@ metadata:
   name: common-service
 spec:
   services:
+  - name: ibm-metering-operator
+    spec:
+      Metering:
```
chenzhiwei: can you change the first letter to lower case? Likely the kubebuilder does not allow upper case ini first letter.

deploy/crds/operator.ibm.com_v1alpha1_metaoperatorset_cr.yaml
```diff
@@ -4,6 +4,9 @@ metadata:
   name: common-service
 spec:
   services:
+  - name: ibm-metering-operator
+    state: present
```
chenzhiwei: please set it to absent, later we will do a bulk update on this flag.

deploy/crds/operator.ibm.com_v1alpha1_metaoperatorconfig_cr.yaml
```diff
@@ -4,6 +4,16 @@ metadata:
...
+        iamNamespace: kube-system
+        mongodb:
+          host: mongodb.kube-system.svc.cluster.local
+      MeteringUI:
+        iamNamespace: kube-system
+        mongodb:
+          host: mongodb.kube-system.svc.cluster.local
```
chenzhiwei: @horis233 @DanielXLee can we set the spec to {}?

All the default values are set inside the CSV file, this can decouple the service operator and meta-operator.

The service operator controls everything and they don't need to notify meta-operator about its change.

```
spec:
  metering: {}
  meteringUI: {}
```

horis233: @chenzhiwei yes, we support to set the value to {}. Then meta operator will just use the alm-example to generate crd.

deploy/crds/operator.ibm.com_v1alpha1_metaoperatorconfig_cr.yaml
sgrube: done

deploy/crds/operator.ibm.com_v1alpha1_metaoperatorset_cr.yaml
sgrube: done

deploy/crds/operator.ibm.com_v1alpha1_metaoperatorconfig_cr.yaml
sgrube: done

PR #81, opened 2020-02-17
enable multi-ns for meta-operator

chenzhiwei:
fixes https://github.com/IBM/meta-operator/issues/50

/cc @gyliu513 @DanielXLee @horis233

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -22,12 +22,13 @@ spec:
           imagePullPolicy: Always
           env:
             - name: WATCH_NAMESPACE
-              valueFrom:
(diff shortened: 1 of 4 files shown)


deploy/operator.yaml
```diff
@@ -22,12 +22,13 @@ spec:
           imagePullPolicy: Always
           env:
             - name: WATCH_NAMESPACE
```
gyliu513: Do we need to add `value: ""`?

https://github.com/operator-framework/operator-sdk/blob/master/doc/operator-scope.md#cluster-scoped-operator-usage

PR #86, opened 2020-02-18
integrate iam operator with meta operator

posriniv:


deploy/crds/operator.ibm.com_v1alpha1_metaoperatorconfig_cr.yaml
```diff
@@ -17,3 +17,6 @@ spec:
   - name: ibm-cert-manager-operator
     spec:
       certManager: {}
+  - name: ibm-iam-operator
+    spec:
+      Authentication: {}
```
chenzhiwei: you need to add all the kinds here, meta-operator only create your operator CR that defined here.

By the way, you'd better lowercase the first letter.

PR #95, opened 2020-02-20
Ensure only one metaoperatorconfig and metaoperatorcatalog in one namespace

### LINKED ISSUE #76: ensure there is only one MetaOperatorCatalog and MetaOperatorConfig in a namespace
the logic is:

Get the list of MetaOperatorCatalog in one namespace, if there are more than one, then log error message and use the first one.

horis233:
fix #76

### DIFF
--- a/pkg/controller/metaoperatorset/metaoperatorset_controller.go
+++ b/pkg/controller/metaoperatorset/metaoperatorset_controller.go
@@ -168,32 +168,11 @@ func (r *ReconcileMetaOperatorSet) Reconcile(request reconcile.Request) (reconci
 	}
 
 	// Fetch the MetaOperatorCatalog instance
-	mocList := &operatorv1alpha1.MetaOperatorCatalogList{}
-	if err := r.client.List(context.TODO(), mocList, &client.ListOptions{Namespace: setInstance.Namespace}); err != nil {
-		if errors.IsNotFound(err) {
-			return reconcile.Result{}, nil
-		}
-		return reconcile.Result{}, err
-	}
+	moc, err := r.listandDeleteCatalog(setInstance.Namespace)
 
-	switch mocnum := len(mocList.Items); {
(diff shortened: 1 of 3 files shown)


DanielXLee:
@horis233 Maybe we can delete the newer config or catalog cr if we find the size >1

pkg/controller/metaoperatorset/metaoperatorset_controller.go
```diff
@@ -160,41 +160,79 @@ func (r *ReconcileMetaOperatorSet) Reconcile(request reconcile.Request) (reconci
...
+	case cscnum < 1:
+		return reconcile.Result{}, nil
+	case cscnum > 1:
+		reqLogger.Error(fmt.Errorf("multiple MetaOperatorConfig in one namespace"),
+		 "There are multiple MetaOperatorConfig custom resource in "+
+		 request.Namespace+
+		 ". Choose the first one "+
+		 cscList.Items[0].Name+
+		 " and Delete the others")
+		 for i:=1; i < cscnum; i++ {
```
gyliu513: Instead of deleting, can we only log a warning message and use the first one? @chenzhiwei @horis233

horis233: Sure, I will log an error message and notify the user meta-operator to pick the first one.

gyliu513: Thanks, I think we should leave the `delete` operations to end user.

horis233:
@gyliu513 done, remove the delete logic and add a notification for users to delete.

pkg/controller/metaoperatorset/metaoperatorset_controller.go
chenzhiwei: yes, we should not touch the users' data.

PR #108, opened 2020-02-23
Onboard ibm-ingress-nginx-operator into meta-operator

liqlin2015:


horis233:
@liqlin2015 Now this operator is in the absent state, so this onboard looks good to me, I will approve it at first, and then you can e2e test with OperatorHub.

From my test, the operator is using a placeholder image.
Screen Shot 2020-02-24 at 9 30 31 AM

PR #173, opened 2020-03-06
Enhance push csv script to support update csv

DanielXLee:

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/common/scripts/push-csv.sh
+++ b/common/scripts/push-csv.sh
@@ -47,13 +47,26 @@ else
 fi
 
 # Push application to repository
-echo "Push package ${QUAY_REPOSITORY} into namespace ${QUAY_NAMESPACE}"
-curl -H "Content-Type: application/json" \
-     -H "Authorization: ${AUTH_TOKEN}" \
-     -XPOST https://quay.io/cnr/api/v1/packages/"${QUAY_NAMESPACE}"/"${QUAY_REPOSITORY}" -d '
-{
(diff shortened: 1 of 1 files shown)


common/scripts/push-csv.sh
```diff
@@ -47,13 +47,26 @@ else
...
+# Delete application release in repository
+function delete_csv() {
+  echo "Delete release ${RELEASE} of package ${QUAY_REPOSITORY} from namespace ${QUAY_NAMESPACE}"
+  curl -H "Content-Type: application/json" \
+      -H "Authorization: ${AUTH_TOKEN}" \
+      -XDELETE https://quay.io/cnr/api/v1/packages/"${QUAY_NAMESPACE}"/"${QUAY_REPOSITORY}"/"${RELEASE}"/helm
+}
 
+#-------------------------------------- Main --------------------------------------#
+delete_csv
```
gyliu513: @DanielXLee so with this fix, we will only have one CSV in the quay.io for ODLM?

DanielXLee: yes, when we want to update the CSV, we don't need to delete the whole ODLM application

gyliu513: Then do we need some flags or options here to disable the `delete_csv`?

DanielXLee: @gyliu513 we don't need a flag for `delete_csv`, I added this feature just to remove the old CSV from quay.io.

PR #193, opened 2020-03-10
Add default handle logic for registryNamespace

### LINKED ISSUE #191: Handle default value for registryNamespace
/kind feature

The `registryNamespace` is an optional property inside of `OperandRequest`, if user don't set this value when creating an `OperandRequest` CR, `OperandRequest` controller will set a default value is the current namespace in which the request is defined

DanielXLee:
The registryNamespace is an optional property inside of OperandRequest, if user don't set this value when creating an OperandRequest CR, OperandRequest controller will set a default value is the current namespace in which the request is defined

Fixes #191

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,7 +111,7 @@ code-dev: ## Run the default dev commands which are the go tidy, fmt, vet then e
 	- make build
 
 run: ## Run against the configured Kubernetes cluster in ~/.kube/config
-	WATCH_NAMESPACE= go run ./cmd/manager/main.go --zap-level=5 --zap-encoder=console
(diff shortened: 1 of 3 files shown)


deploy/crds/operator.ibm.com_v1alpha1_operandrequest_cr.yaml
```diff
@@ -6,7 +6,6 @@ metadata:
 spec:
   requests:
   - registry: common-service
-    registryNamespace: ibm-common-services
```
chenzhiwei: I think we can leave this in the CR, this can help others understand this CR.

DanielXLee: Okay, add it back

PR #201, opened 2020-03-11
Operandconfig delete

### LINKED ISSUE #199: Enhance OperandConfig to delete CR when they are removed from the instance
/kind feature

1.  Add label when creating CR
2. Check the CR in the `alm-example` but not in the `operandConfig`.
3. If found the Cr with the label but not in the `operandConfig`, delete it.

horis233:

Fixes #199 

This PR is used to implement the function when the CR is removed from the operandconfig ODLM can remove the CR from the cluster.
 
Moreover, expose more error messages in the logs.
  
**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/operandrequest/reconcile_operand.go
+++ b/pkg/controller/operandrequest/reconcile_operand.go
@@ -34,7 +34,7 @@ import (
 	util "github.com/IBM/operand-deployment-lifecycle-manager/pkg/util"
 )
 
-const t string = "true"
+const t = "true"
 
 func (r *ReconcileOperandRequest) reconcileOperand(requestInstance *operatorv1alpha1.OperandRequest) *multiErr {
 	klog.V(1).Info("Reconciling Operand")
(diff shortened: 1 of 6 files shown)


pkg/controller/operandrequest/reconcile_operand.go
```diff
@@ -153,70 +161,113 @@ func (r *ReconcileOperandRequest) createUpdateCr(service *operatorv1alpha1.Confi
...
 						continue
 					}
-					existingCR.Object["spec"] = unstruct.Object["spec"]
-					if crUpdateErr := r.client.Update(context.TODO(), existingCR); crUpdateErr != nil {
-						stateUpdateErr := r.updateServiceStatus(csc, service.Name, crdName, operatorv1alpha1.ServiceFailed)
+					if existingCR.Object["metadata"].(map[string]interface{})["labels"] != nil && existingCR.Object["metadata"].(map[string]interface{})["labels"].(map[string]interface{})["operator.ibm.com/opreq-control"] == t {
+						existingCR.Object["spec"] = unstruct.Object["spec"]
+						if crUpdateErr := r.client.Update(context.TODO(), &existingCR); crUpdateErr != nil {
+							stateUpdateErr := r.updateServiceStatus(csc, service.Name, crdName, operatorv1alpha1.ServiceFailed)
+							if stateUpdateErr != nil {
```
chenzhiwei: code looks good, but it is better to separate the nested `if/for` into smaller functions.

Linux kernel code style: https://github.com/torvalds/linux/blob/master/Documentation/process/coding-style.rst#7-centralized-exiting-of-functions

> The answer to that is that if you need more than 3 levels of indentation, you're screwed anyway, and should fix your program.

Let's don't use more than 5 nested levels of `for/if`, the current nested level is 6. The less the better.

We can merge the code for now, and later enhance the logic.

horis233: @chenzhiwei Yes, I plan to reflect the code in the reconcile operand, Currently, it is unreadable.

PR #223, opened 2020-03-16
Update iam config

### LINKED ISSUE #222: After remove iam from request, not all the config status return to ready.
/kind bug

[A clear and concise description of what the bug is.]

After installation and remove the IAM operator from operandRequest and operandConfig is not in the correct status.

Screen Shot 2020-03-16 at 4 30 22 PM

All the custom resources are in the ready status.

**Environment:**

- ODLM version:
- Minikube/KIND/OCP version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

horis233:

All the custom resource names in the operandconfig instance should be the kind name of custom resource definition with the lower case first letter.
Otherwise, there will be a defect when updating the operandconfig status.
In the PR is used to update the custom resource format. 

Fixes #222 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/deploy/crds/operator.ibm.com_v1alpha1_operandconfig_cr.yaml
+++ b/deploy/crds/operator.ibm.com_v1alpha1_operandconfig_cr.yaml
@@ -10,7 +10,7 @@ spec:
       meteringUI: {}
   - name: ibm-licensing-operator
     spec:
-      IBMLicensing: {}
+      iBMLicensing: {}
   - name: ibm-mongodb-operator
     spec:
       mongoDB: {}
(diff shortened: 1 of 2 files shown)


deploy/crds/operator.ibm.com_v1alpha1_operandconfig_cr.yaml
```diff
@@ -10,7 +10,7 @@ spec:
       meteringUI: {}
   - name: ibm-licensing-operator
     spec:
-      IBMLicensing: {}
+      iBMLicensing: {}
```
DanielXLee: @horis233 why change the name, if these names changed, we also need to update cicd test guide, https://github.ibm.com/IBMPrivateCloud/roadmap/blob/master/feature-specs/cicd/common-service-1Q-e2e-testing.md

horis233: @DanielXLee Because when deleting a custom resource from the OperandConfig, ODLM can get custom name from the `kind` in the `alm-example`.  When updating status, I set the logic to delete the custom resource with the name as `kind` only lower its first letter. If we don't set this naming rule, we need to use a loop to check the CRs in the operator. 

If you think it is feasible, I will update the e2e testing in the roadmap afterwards.

horis233:
Based on the discussion with @DanielXLee on slack, I will close this PR and update status update logic in the ODLM

PR #240, opened 2020-03-20
fix typo in design doc

hmtai:

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/docs/design/operand-deployment-lifecycle-manager.md
+++ b/docs/design/operand-deployment-lifecycle-manager.md
@@ -177,9 +177,7 @@ Suppose IAM Operator has two CRDs: Apikey and Identity:
                 "name": "iam-identity"
               },
               "spec": {
-                "key": "value",
-                "nested": {
-                  "key": "value"
+                "key": "value"
                 }
               }
             }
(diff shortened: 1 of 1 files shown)


docs/design/operand-deployment-lifecycle-manager.md
```diff
@@ -177,9 +177,7 @@ Suppose IAM Operator has two CRDs: Apikey and Identity:
                 "name": "iam-identity"
               },
               "spec": {
-                "key": "value",
-                "nested": {
```
gyliu513: why kill here?

hmtai: The example in step 1 just has the "key": "value" like this for identity:
```
- name: iam
  spec:
    apikey:
      key: value
      nested:
        key: value
    identity:
      key: value
```
Not sure my understanding is correct,in the csv it should like this:
```
        {
          "apiVersion": "iam.operator.ibm.com/v1alpha1",
          "kind": "Identity",
          "metadata": {
            "name": "iam-identity"
          },
          "spec": {
            "key": "value"
          }
        }
```

gyliu513: Good catch, thanks.

PR #261, opened 2020-03-26
changed logging channel

dian-xh:
fixes 36530


DanielXLee:
@dian-xh when the cr changed, you also need to update the ODLM CSV file. https://github.com/IBM/operand-deployment-lifecycle-manager/blob/bfbdc8de62aa002b6950afdd7243ae8ee06234c6/deploy/olm-catalog/operand-deployment-lifecycle-manager/1.1.0/operand-deployment-lifecycle-manager.v1.1.0.clusterserviceversion.yaml#L302

PR #286, opened 2020-03-31
Add resource annotation and enable all common services

horis233:

1. Enable all the services
2. Add resources annotation for each crd

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/apis/operator/v1alpha1/operandrequest_types.go
+++ b/pkg/apis/operator/v1alpha1/operandrequest_types.go
@@ -171,6 +171,20 @@ type MemberStatus struct {
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:path=operandrequests,shortName=opreq,scope=Namespaced
 // +operator-sdk:gen-csv:customresourcedefinitions.displayName="OperandRequest"
+// +operator-sdk:gen-csv:customresourcedefinitions.resources=`Namespace,v1,""`
+// +operator-sdk:gen-csv:customresourcedefinitions.resources=`Deployment,v1,""`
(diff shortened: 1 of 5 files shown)


pkg/apis/operator/v1alpha1/operandrequest_types.go
```diff
@@ -171,6 +171,20 @@ type MemberStatus struct {
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:path=operandrequests,shortName=opreq,scope=Namespaced
 // +operator-sdk:gen-csv:customresourcedefinitions.displayName="OperandRequest"
+// +operator-sdk:gen-csv:customresourcedefinitions.resources=`Namespace,v1,""`
+// +operator-sdk:gen-csv:customresourcedefinitions.resources=`Deployment,v1,""`
```
gyliu513, line 175: @horis233 one question, can you show some detail for what is the use of those comments? Thanks.

horis233: @gyliu513  These annotations are used to generate the `resources` field in the csv:

```
    - description: OperandConfig is the Schema for the operandconfigs API
      displayName: OperandConfig
      kind: OperandConfig
      name: operandconfigs.operator.ibm.com
      resources:          <------ this section
      - kind: Catalogsource
        name: ""
        version: v1alpha1
      - kind: Clusterserviceversion
        name: ""
        version: v1alpha1
      - kind: Configmap
        name: ""
        version: v1
      - kind: Deployment
        name: ""
        version: v1
      - kind: Installplan
```

The `resources` fields are required in the RH certification. If without these annotations, every time when the CSV is re-generated, the `resources` fields will be removed. Developers need to manually add them back.

PR #327, opened 2020-04-14
Remove useless code from main.go

DanielXLee:

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -18,40 +18,26 @@ package main
 
 import (
 	"context"
-	"errors"
 	"flag"
-	"fmt"
 	"os"
 	"runtime"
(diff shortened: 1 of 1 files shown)


chenzhiwei:
The Metrics is generated by operator-sdk automatically and seems most operators have it.

Why removing it?

DanielXLee:
@chenzhiwei This code only used by Prometheus operator, but we don't need to report any for data for this operator and don't depend on this operator
https://github.com/openshift/ocs-operator/blob/9f813037e154084af1bacfc86aaafee89e32a4c2/cmd/manager/main.go#L148
This is an official operator of openshift, they also removed these useless codes.

PR #349, opened 2020-04-20
Skip reconcile bindinfo when requestNamespaces less than 2

DanielXLee:

If the `requestNamespaces` less than 2, means there is one request or not.
If not, we certainly do not have to
If there is one, the request namespace must be equal to the operand namespace, we also don't copy ....

Fixes #

**Release note**:

```release-note

```


horis233:
@DanielXLee It doesn't work for most of the Cloud Pak scenario. 

When a cloudpak requests common services, it will create an OperandRequest in its own namespace and common services operands are deployed in `ibm-common-services` namespace. We need use OperandBindInfo to copy secret and configmap to the cloudpak namespace.

PR #368, opened 2020-04-28
Add a stable odlm api version v1

DanielXLee:
Add a new apiversion `v1` in CRD, current we can support 2 apiversion `v1` and `v1alpha1`

Fixes #

**Release note**:

```release-note

```


chenzhiwei:
@horis233 please check if this is fine to upgrade to new apiversion v1.

DanielXLee:
@gyliu513 Because the dev sprint is over, add v1 api will generate too many code change, maybe have some potential risks, so just add a v1 version into CRD, do you think we need add a real `v1` API?

DanielXLee:
If want to migrate the existing CR instance to the new API version, it need some manual steps:
https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version

deploy/crds/operator.ibm.com_operandbindinfos_crd.yaml
```diff
@@ -92,8 +92,18 @@ spec:
               type: array
           type: object
       type: object
-  version: v1alpha1
```
chenzhiwei, line 95: set `version: v1`.

PR #388, opened 2020-05-16
create beta channel and update tag to digest 

horis233:

- Create a beta channel
- Update tag to digest 

**Release note**:

```release-note
create the beta channel
```

### DIFF
--- a/deploy/olm-catalog/operand-deployment-lifecycle-manager/1.1.0/operand-deployment-lifecycle-manager.v1.1.0.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/operand-deployment-lifecycle-manager/1.1.0/operand-deployment-lifecycle-manager.v1.1.0.clusterserviceversion.yaml
@@ -62,8 +62,7 @@ metadata:
                 "name": "ibm-commonui-operator",
                 "spec": {
                   "commonWebUI": {},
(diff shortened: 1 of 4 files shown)


deploy/olm-catalog/operand-deployment-lifecycle-manager/1.1.0/operand-deployment-lifecycle-manager.v1.1.0.clusterserviceversion.yaml
```diff
@@ -62,8 +62,7 @@ metadata:
                 "name": "ibm-commonui-operator",
                 "spec": {
                   "commonWebUI": {},
-                  "legacyHeader": {},
-                  "navconfiguration": {}
+                  "legacyHeader": {}
```
chenzhiwei, line 65: why doing this change? the v1.1.0 is in stable-v1 channel, we should not update it.

And this change does not impact the 3.3 release as we use yaml file to create the operandconfig.

Please rollback this change.

horis233: I roll it back

PR #391, opened 2020-05-21
continue the loop when can't get request

horis233:

It is a follow-up PR for the #390. It is to cover the case that there is an error when fetching the operandrequest. 

**Release note**:

```release-note
NONE
```


renshichuang:
Approved.

PR #428, opened 2020-06-18
Enhance request member status

### LINKED ISSUE #427: Enhance request member status update
/kind feature

1. Set member status after operator and operand reconcile
2. Move registry status updates from request controller to registry controller

DanielXLee:
1. Set member status after operator and operand reconcile
2. Move registry status updates from request controller to registry controller
3. Comment some ut and e2e tests, it will be updated in a new pr

Fixes #427 

**Release note**:

```release-note

```


pkg/apis/operator/v1alpha1/operandregistry_types.go
```diff
@@ -86,7 +86,7 @@ type OperandRegistryStatus struct {
 	// Phase describes the overall phase of operators in the OperandRegistry
 	// +operator-sdk:gen-csv:customresourcedefinitions.statusDescriptors=true
 	// +optional
-	Phase OperatorPhase `json:"phase,omitempty"`
+	Phase RegistryPhase `json:"phase,omitempty"`
```
chenzhiwei, line 89: don't change the api types, let's keep it as it.

When we bump the new apiversion, we can do this, but still need to keep backward compatibility.

DanielXLee: It doesn't matter, API not change, `OperatorPhase` and `RegistryPhase` are both `string` type, so no impact.

Issue #426, opened 2020-06-17
Add a installMode parameter into the OperandRegistry


horis233:
/kind feature

Users want to deploy operator with different install modes including single namespace, multiple namespaces and all namespaces and can decide to deploy CR into which namespace.

Therefore an `installMode` parameter is required to be set into the OperandRegistry for each operator.

- All namespaces:
Add an `installmode` parameter into the OperandRegistry, if it values is `AllNamespaces`, then the operator will be deployed into `openshift-operators` namespace and the user can set the namespace field in the OperandRegistry to decide which namespace the CR will be deployed in
```
        - name: ibm-metering-operator
          installMode: allnamespaces 
          namespace: <the namespace of CR>
          channel: stable-v1
          packageName: ibm-metering-operator-app
          scope: public
          sourceName: opencloud-operators
          sourceNamespace: openshift-marketplace
```

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

PR #429, opened 2020-06-19
support operator install mode: cluster and namespace

chenzhiwei:
fixes https://github.com/IBM/operand-deployment-lifecycle-manager/issues/426

I tested on OpenShift 4.3, the new CRD will be always applied on the cluster even if the apiversion is not changed.


pkg/controller/operandrequest/reconcile_operator.go
```diff
@@ -225,28 +238,26 @@ func (r *ReconcileOperandRequest) deleteSubscription(operandName string, request
 		}
 	}
 
-	opt := registryInstance.GetOperator(operandName)
```
chenzhiwei, line 228: @DanielXLee @horis233  Please check here whether we really need this or not.

chenzhiwei:
Manually build a catalogsource image and tested the code, it works as expected.

But when remove etcd and re-enable etcd, it will encounter error: https://github.com/operator-framework/operator-lifecycle-manager/issues/1570

PR #432, opened 2020-06-19
Update finializer management

### LINKED ISSUE #431: OperandRequest finalizer missing if there're finalizers already existed but not added by ODLM
/kind bug

ODML has its own finalizer `finalizer.request.ibm.com` to control OperandRequest lifecycle.

We created OperandRequest and added custom finalizer `ibm-management-installer.management.ibm.com/finalizer` together w/ ODML to control the OperandRequest lifecycle at our own pace. Kubernetes does allow to add multiple finalizers on a certain resource.

However, after above steps are done, when we check the OperandRequest, it only has the custom finalizer:
```
$ oc get opreq -n cp4m ibm-management-foundation -o yaml
kind: OperandRequest
metadata:
  finalizers:
  - ibm-management-installer.management.ibm.com/finalizer
```

If we do not add custom finalizer, the ODML finalizer will be there as expected:
```
$ oc get opreq -n ibm-common-services catalog-ui-request -o yaml
kind: OperandRequest
metadata:
  finalizers:
  - finalizer.request.ibm.com
```

This behavior leads to an issue when we want to use custom finalizer together w/ ODML finalizer.

When delete OperandRequest along w/ custom finalizer, ODML will lost the control of  OperandRequest completely due to missing ODML finalizer, so that no chance for ODML to do some clean up, e.g. to delete Operand CR, hence, leave all the corresponding resources, e.g. pods, remain even OperandRequest has gone.

Either ODML finalizer needs to be added no matter custom finalizer used or not, or ODML needs to clearly tell users not to use custom finalizer, by code or documentation.

By reading ODML code, I learned here's the problem comes:

https://github.com/IBM/operand-deployment-lifecycle-manager/blob/0e5353ae981a58b048fdfbf501b78d44c78f083f/pkg/controller/operandrequest/operandrequest_controller.go#L150-L154

The code checks if there're finalizers already existed, no matter whether they're added by ODML or not, it will never add ODML default finalizer.

If we allow multiple finalizers, usually, we will do the check safely like this:
```
if !utils.ContainsString(existingResource.Finalizers, customFinalizer) {
    currentResource.Finalizers = append(existingResource.Finalizers, customFinalizer)
}
```

Or, we should clearly deny the use of custom finalizer in code, e.g. throw error when the code detects there're finalizers already existed but not added by ODML, or, documented in spec.

Currently, it's not clear in this area which I believe very easy to be hit when more users start to use ODML.

horis233:

This PR enhances the logic for the finalizer management. 
Add `ensureFinalizer` function to make sure the specific finalizer exists.
Add `removeFinalizer` function to remove a specific finalizer.

Therefore, we can keep existing finalizers added by end-users.

Fixes #431

**Release note**:

```release-note

```

### DIFF
--- a/pkg/controller/operandrequest/operandrequest_controller.go
+++ b/pkg/controller/operandrequest/operandrequest_controller.go
@@ -175,11 +175,8 @@ func (r *ReconcileOperandRequest) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
 	}
 
-	// Add finalizer
-	if requestInstance.GetFinalizers() == nil {
-		if err := r.addFinalizer(requestInstance); err != nil {
-			return reconcile.Result{}, err
-		}
+	if err := r.addFinalizer(requestInstance); err != nil {
+		return reconcile.Result{}, err
 	}
 
 	// Remove finalizer when DeletionTimestamp none zero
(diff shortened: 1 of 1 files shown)


morningspace:
@horis233 Thanks for your quick response :+1:

Issue #435, opened 2020-06-22
After deleting the OperandRequest, operators aren't be uninstalled


horis233:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

When the OperandRequest is deleted, ODLM won't uninstall the relevant operators from the cluster.

The reason for this issue is the delete logic depends on the OperandRegistry status.

https://github.com/IBM/operand-deployment-lifecycle-manager/blob/master/pkg/controller/operandrequest/reconcile_operator.go#L193

Since the OperandRegistry status updating is decoupled from the OperandRequest reconciliation, there is no way to set the order for OperandRequest reconciliation and OperandRegistry reconciliation.

Therefore, there is a race condition between OperandRequest and OperandRegistry, which will cause operators can be deleted.

**What did you expect to happen:**

When the OperandRequest is removed, all the relevant operators will be uninstalled.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- ODLM version:
- Minikube/KIND/OCP version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

horis233:
fixed by #438

PR #438, opened 2020-06-23
Update operator deletion check

### LINKED ISSUE #434: Can't Delete Operators from the OperandRequest
/kind bug

[A clear and concise description of what the bug is.]

When users remove Operators from an OperandRequest, ODLM can't delete the Operator from the cluster.

The reason for this issue is the delete logic depends on the OperandRegistry status.

https://github.com/IBM/operand-deployment-lifecycle-manager/blob/master/pkg/controller/operandrequest/reconcile_operator.go#L108

Since we decouple the OperandRegistry status update from the OperandRequest reconciliation, there is no way to set the order for OperandRequest reconciliation and OperandRegistry reconciliation. Therefore, there is a race condition between OperandRequest and OperandRegistry, which will cause the operator can be deleted.
 
When the operator is removed from OperandRequest, the operator should be uninstalled.

**Environment:**

- ODLM version:
- Minikube/KIND/OCP version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

horis233:

This PR is for decoupling the OperandRegistry reconciliation with the OperandRegistry to avoid the race condition.
Then it can fix the deletion detects mentioned in #434 and #435.
 
Fixes #434 #435 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/operandrequest/reconcile_operator.go
+++ b/pkg/controller/operandrequest/reconcile_operator.go
@@ -29,12 +29,12 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	operatorv1alpha1 "github.com/IBM/operand-deployment-lifecycle-manager/pkg/apis/operator/v1alpha1"
 )
 
-func (r *ReconcileOperandRequest) reconcileOperator(requestInstance *operatorv1alpha1.OperandRequest, reconcileReq reconcile.Request) error {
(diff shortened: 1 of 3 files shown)


pkg/controller/operandrequest/reconcile_operator.go
```diff
@@ -305,36 +255,46 @@ func (r *ReconcileOperandRequest) deleteSubscription(operandName string, request
...
+func (r *ReconcileOperandRequest) getNeedDeletedOperands(requestInstance *operatorv1alpha1.OperandRequest) (gset.Set, error) {
 	klog.V(3).Info("Getting the operater need to be delete")
-	requestOperands := gset.NewSet()
-	for _, req := range requestInstance.Spec.Requests {
-		for _, o := range req.Operands {
-			requestOperands.Add(o.Name)
-		}
+	deployedOperands := gset.NewSet()
+	for _, req := range requestInstance.Status.Members {
+		deployedOperands.Add(req.Name)
```
DanielXLee, line 262: This logic cannot cover multi-request share the same operator

PR #449, opened 2020-06-29
Add finalizers to operandRegistry and operandConfig

### LINKED ISSUE #395: Add finializer for OperandConfig and OperandRequest
/kind feature

If a user accidentally delete OperandConfig or OperandRequest, it will break the uninstall of the OperandRequest

horis233:

Fixes #395

This PR set finalized to OperandConfig and OperandRegistry, in case they are deleted by users accidentally before operandRequests are deleted.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/controller/operandrequest/reconcile_operand.go
+++ b/pkg/controller/operandrequest/reconcile_operand.go
@@ -104,7 +104,6 @@ func (r *ReconcileOperandRequest) reconcileOperand(requestInstance *operatorv1al
 			if csv.Status.Phase != olmv1alpha1.CSVPhaseSucceeded {
 				klog.Errorf("The ClusterServiceVersion for Subscription %s is not Ready", opdConfig.Name)
 				requestInstance.SetMemberStatus(operand.Name, operatorv1alpha1.OperatorInstalling, "")
-				continue
 			}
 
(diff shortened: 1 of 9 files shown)


pkg/controller/operandrequest/reconcile_operand.go
```diff
@@ -104,7 +104,6 @@ func (r *ReconcileOperandRequest) reconcileOperand(requestInstance *operatorv1al
 			if csv.Status.Phase != olmv1alpha1.CSVPhaseSucceeded {
 				klog.Errorf("The ClusterServiceVersion for Subscription %s is not Ready", opdConfig.Name)
 				requestInstance.SetMemberStatus(operand.Name, operatorv1alpha1.OperatorInstalling, "")
-				continue
```
horis233, line 107: @DanielXLee I remove this continue is because of ibm-iam-operator, which CSV won't be in the succeeded status until its dependency operands are deployed.

DanielXLee: I think we need to create an issue for iam update their operator don't depend on other operators in their operator deployment.

DanielXLee: okay, remove it at first, after IAM operator enhanced. add it back again

PR #452, opened 2020-07-03
Check catalog source status

### LINKED ISSUE #448: Check CatalogSource status in OperandRegistry controller to make sure the registry phase
/kind feature

DanielXLee:

Fixes #448

**Release note**:

```release-note

```

### DIFF
--- a/pkg/apis/operator/v1alpha1/operandregistry_types.go
+++ b/pkg/apis/operator/v1alpha1/operandregistry_types.go
@@ -17,6 +17,7 @@
 package v1alpha1
 
 import (
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -98,6 +99,11 @@ type OperandRegistryStatus struct {
 	// OperatorsStatus defines operators status and the number of reconcile request
 	// +optional
 	OperatorsStatus map[string]OperatorStatus `json:"operatorsStatus,omitempty"`
+	// Conditions represents the current state of the Request Service
+	// +optional
+	// +listType=set
+	// +operator-sdk:gen-csv:customresourcedefinitions.statusDescriptors=true
+	Conditions []Condition `json:"conditions,omitempty"`
 }
 
 // OperatorStatus defines operators status and the number of reconcile request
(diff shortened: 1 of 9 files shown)


pkg/controller/operandregistry/operandregistry_controller.go
```diff
@@ -51,6 +54,11 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
+	// Add extenal api scheme
```
horis233, line 57: @DanielXLee Do we need to add the external api scheme here?
I think https://github.com/IBM/operand-deployment-lifecycle-manager/blob/master/pkg/apis/addtoscheme_operator_v1alpha1.go#L28 is enough to use OLM api.

pkg/controller/operandregistry/operandregistry_controller.go
```diff
@@ -121,28 +129,8 @@ func (r *ReconcileOperandRegistry) Reconcile(request reconcile.Request) (reconci
...
-		instance.UpdateRegistryPhase(operatorv1alpha1.RegistryInit)
-	} else {
-		instance.UpdateRegistryPhase(operatorv1alpha1.RegistryRunning)
-	}
-	if err := r.client.Status().Update(context.TODO(), instance); err != nil {
-		return reconcile.Result{}, err
-	}
-
 	// Set Finalizer for the OperandRegistry
-	added := instance.EnsureFinalizer()
```
horis233, line 144: I wrote the condition in this way just want to make the code more readable.
I am ok with the change.
Could we enhance the comment to make it more clear?
For example
```
// // Set Finalizer for the OperandRegistry. If the OperandRegistry finalizer is added to the finalizer list, EnsureFinalizer() will return true. If the OperandRegistry finalizer already exists, EnsureFinalizer() will return false.
```

DanielXLee: Updated

pkg/controller/operandregistry/operandregistry_controller.go
DanielXLee: you are right, I'll remove it

pkg/apis/operator/v1alpha1/operandregistry_types.go
```diff
@@ -240,8 +246,25 @@ func (r *OperandRegistry) GetAllReconcileRequest() []reconcile.Request {
...
 
-func init() {
-	SchemeBuilder.Register(&OperandRegistry{}, &OperandRegistryList{})
+// SetReadyCondition creates a Condition to claim Ready
+func (r *OperandRegistry) SetReadyCondition(name string, rt ResourceType, cs corev1.ConditionStatus) {
+	c := newCondition(ConditionReady, cs, string(rt)+" is ready", string(rt)+" "+name+" is ready")
+	r.setCondition(*c)
+}
+
+// SetNoneCondition creates a Condition to claim Ready
```
chenzhiwei, line 255: line 249 and 255 are same, is this the expected?

DanielXLee: These are 2 different condition setup method, seems the comment need update

PR #460, opened 2020-07-06
Check the operator generated from the OCP console

### LINKED ISSUE #457: Add label to prevent subscription and CR uninstall
/kind feature

If users don't want ODLM to delete the subscription and/or CR, they can add the following label to prevent subscription and CR uninstallation.
```
labels:
  operator.ibm.com/opreq-do-not-uninstall: 'true'
```

horis233:
This PR Implement two main functions:

- Subscriptions (operators) generated from the OCP console use the operator package name as the subscription name.
ODLM will check if the subscription/operator deployed by OCP console, then it won't deploy the same operator in the same namespace.

- If the subscription deployed by OCP console has the label "operator.ibm.com/opreq-control", then ODLM should manage CRs for the operator.

Fixes #457 

**Release note**:

```release-note

```

### DIFF
--- a/pkg/controller/operandrequest/reconcile_operand.go
+++ b/pkg/controller/operandrequest/reconcile_operand.go
@@ -135,21 +135,8 @@ func (r *ReconcileOperandRequest) reconcileOperand(requestKey types.NamespacedNa
 // getCSV retrieves the ClusterServiceVersion
 func (r *ReconcileOperandRequest) getClusterServiceVersion(subName, packageName, subNamespace string) (*olmv1alpha1.ClusterServiceVersion, error) {
 	klog.V(3).Infof("Looking for the ClusterServiceVersion for Subscription %s in the namespace %s", subName, subNamespace)
-	sub, err := r.olmClient.OperatorsV1alpha1().Subscriptions(subNamespace).Get(subName, metav1.GetOptions{})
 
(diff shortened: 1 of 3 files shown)


pkg/controller/operandrequest/reconcile_operand.go
```diff
@@ -134,21 +134,32 @@ func (r *ReconcileOperandRequest) reconcileOperand(requestKey types.NamespacedNa
...
+		if err != nil && !errors.IsNotFound(err) {
+			klog.Error("Failed to list Subscriptions: ", err)
+			return nil, err
+		}
 	}
-	if err != nil {
-		klog.Error("Failed to list Subscriptions: ", err)
-		return nil, err
+
+	if errors.IsNotFound(err) {
```
DanielXLee, line 155: Seems this `errors.IsNotfound` duplicate with above

horis233: The first `errors.IsNotfound` is for checking the subscription with `operator.name`. The second one is for checking the subscription with the operator package name.

DanielXLee: Looks a bit confused, can you create a method to get the sub with `operator.name` and package name?

horis233: Yes, sure

PR #477, opened 2020-07-18
Delete copies when the bindinfo is deleted

### LINKED ISSUE #476: When the operandbindinfo get deleted, the secret/configmap copies should be removed as well.
/kind feature

When the OperandBindInfo gets deleted, the secret/configmap copies should be removed as well.

The deletion of OperandBindInfo happens when the operator deleted from the OperandRequest. It means the API info provided in the secret/configmap copies are not valid as more.

Moreover, the deletion of the secret/configmap copies helps to clear up the environment during the uninstall.

horis233:

When the OperandBindInfo instance is deleted, the secret and configmap copies owned by this instance will be deleted as well.

Fixes #476 

**Release note**:

```release-note

```

### DIFF
--- a/pkg/controller/operandbindinfo/operandbindinfo_controller_test.go
+++ b/pkg/controller/operandbindinfo/operandbindinfo_controller_test.go
@@ -81,6 +81,27 @@ func initReconcile(t *testing.T, r ReconcileOperandBindInfo, req reconcile.Reque
 	secret2 := &corev1.Secret{}
 	err = r.client.Get(context.TODO(), types.NamespacedName{Name: "secret2", Namespace: requestNamespace}, secret2)
 	assert.True(errors.IsNotFound(err))
+
+	bindinfoInstance := &v1alpha1.OperandBindInfo{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: "ibm-operators-bindinfo", Namespace: "ibm-operators"}, bindinfoInstance)
+	assert.NoError(err)
+
+	bindinfoInstance.DeletionTimestamp = now()
+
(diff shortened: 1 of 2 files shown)


pkg/controller/operandbindinfo/operandbindinfo_controller_test.go
```diff
@@ -81,6 +81,27 @@ func initReconcile(t *testing.T, r ReconcileOperandBindInfo, req reconcile.Reque
...
+	res, err = r.Reconcile(req)
+	if res.Requeue {
+		t.Error("Reconcile requeued request as not expected")
+	}
+	assert.NoError(err)
+
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: "cm3", Namespace: requestNamespace}, configmap1)
+	assert.True(errors.IsNotFound(err))
+
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: "secret3", Namespace: requestNamespace}, secret1)
```
Daniel-Fan, line 103: May I ask that why variables configmap1 and secrec1 are with Name cm3 and secret3 in line 70, 78 and here? 
Since cm3 and secret3 are both from [OperandRequest Obj](https://github.com/IBM/operand-deployment-lifecycle-manager/blob/a9576473e19d3173ca1bf88c37fadb2ed19b177b/pkg/controller/operandbindinfo/operandbindinfo_controller_test.go#L199)

horis233: The bindinfo controller will copy the specified secret/configmap to the namespace of OperandRequest and rename them according to the binding section the OperandRequest.

There the secret and configmap in the ut will be renamed to Name secret3 and cm3.

PR #482, opened 2020-08-04
Using internal pipeline

horis233:

CICD Changes 

# Integrating Operator Images

For operator images, the builds should be able to remain largely unchanged, we just need to change where the images are pushed once they are built, and pull in the credentials needed to push to Artifactory.  We are also going to go back to pushing with the final image version tag since we are pushing to Artifactory first, CICD will maintain multiple image versions in our test repo to avoid breaking builds by tagging images both with the release version tag, and with a datestamp tag when they are newly promoted.

There are basically 3 updates that need to happen in your operator repo:

#### 1. Update the `common/scripts/config_docker.sh` script to get Artifactory credentials and log into artifactory.
There should be an existing section in this file that looks like this:
```
DOCKER_REGISTRY="quay.io"
DOCKER_USERNAME="multicloudlab"
DOCKER_PASSWORD=$(${KUBECTL} -n default get secret quay-cred -o jsonpath='{.data.password}' | base64 --decode)
```
You will instead replace it with these lines:
```
DOCKER_REGISTRY="hyc-cloud-private-integration-docker-local.artifactory.swg-devops.com"
DOCKER_USERNAME=$(${KUBECTL} -n default get secret artifactory-cred -o jsonpath='{.data.username}' | base64 --decode)
DOCKER_PASSWORD=$(${KUBECTL} -n default get secret artifactory-cred -o jsonpath='{.data.password}' | base64 --decode)
```

#### 2. Update the registry defined in your `Makefile` 
There is a line in the `Makefile` that sets `quay.io` as the registry to push images to:
```
REGISTRY ?= quay.io/opencloudio
```
You will instead replace it with integration:
```
REGISTRY ?= "hyc-cloud-private-integration-docker-local.artifactory.swg-devops.com/ibmcom"
```

#### 3. Update the `VERSION` variable defined in your Makefile
Most operators are setting a `VERSION` variable using the git commit, and pushing the image tagged with that value using a statement similar to this:
```
VERSION ?= $(shell git describe --exact-match 2> /dev/null || \
                 git describe --match=$(git rev-parse --short=8 HEAD) --always --dirty --abbrev=8)
```
For go operators, you can use this line instead to fetch the `Version` variable from your version.go (the same place cicd will fetch it to decide what tag to pick up, that way you only have to update one location to switch to a new image tag):
```
VERSION ?= $(shell cat ./version/version.go | grep "Version =" | awk '{ print $$3}' | tr -d '"')
```

Fixes #

**Release note**:

```release-note

```


horis233:
@morvencao 
Could you please take a look at the test job?

It reports two issues:

1. `make: command: Command not found`, we use `$(shell command -v kubectl)` in the Makefile, but seems PROW image doesn't have `command`

2. `go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.3.0: sigs.k8s.io/controller-tools@v0.3.0: verifying module: invalid GOSUMDB: malformed verifier id`

horis233:
@morvencao 
Thanks, I will avoid downloading these tool in the prow job.

horis233:
@morvencao Since the failed job returns
 
```
make: command: Command not found
```

Should we use `command -v` to check if a tool exists in our Makefile?

morvencao:
I just updated the branch rules because @DanielXLee required me to remove the build-xxx job from the prow job list for odlm.

horis233:
Thanks @morvencao

@DanielXLee 
I temperately comment out the tasks in `make test` in order to unblock the prow job. Currently, the kubebuilder doesn't support in the s390 architecture.

PR #507, opened 2020-08-28
Add a new fetch csv method for delete

DanielXLee:

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/controllers/common/fetch.go
+++ b/controllers/common/fetch.go
@@ -18,10 +18,14 @@ package common
 
 import (
 	"context"
+	"strings"
+	"time"
 
 	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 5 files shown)


controllers/reconcile_operator.go
```diff
@@ -192,7 +192,11 @@ func (r *OperandRequestReconciler) deleteSubscription(operandName string, reques
 		return nil
 	}
 
-	csv, err := fetch.FetchClusterServiceVersion(r.Client, sub)
+	if sub.Status.CurrentCSV == "" {
```
horis233, line 195: If the `CurrentCSV` is empty, which could be caused by some OLM error. Consider this situation, OLM can't generate Installplan or CSV for the operand.

ODLM should be able to delete the subscription as well. Thus I think we can delete the subscription only if the `CurrentCSV` is empty

DanielXLee: In upgrade process, maybe the CurrentCSV will disappear for a while, if we only delete sub, for the next reconcile, we will not delete the custom resource and CSV

controllers/common/fetch.go
```diff
@@ -115,57 +119,69 @@ func FetchSubscription(c client.Client, name, namespace string, packageName ...s
...
 
-	if sub.Status.Install == nil || sub.Status.InstallPlanRef.Name == "" {
+	if sub.Status.InstallPlanRef != nil {
+		ip, err := FetchInstallPlan(c, sub.Status.InstallPlanRef.Name, sub.Status.InstallPlanRef.Namespace)
+		if err != nil {
+			return nil, err
+		}
+		for _, step := range ip.Status.Plan {
+			if step.Resource.Kind == "ClusterServiceVersion" && strings.HasPrefix(step.Resource.Name, sub.Name) {
+				if step.Status == olmv1alpha1.StepStatusPresent {
```
horis233, line 132: @DanielXLee When I fresh install, I found when the CSV is created, the status of the installplan is `olmv1alpha1.StepStatusCreated` instead of `olmv1alpha1.StepStatusPresent`.

controllers/common/fetch.go
```diff
@@ -115,57 +119,69 @@ func FetchSubscription(c client.Client, name, namespace string, packageName ...s
...
+	}
+	if err := getObjectWithRetry(c, installPlanKey, installPlan); err != nil {
+		return nil, err
+	}
+	return installPlan, nil
+}
+
+// Get object with retry
+func getObjectWithRetry(c client.Client, key types.NamespacedName, obj runtime.Object) error {
+	if err := wait.PollImmediate(time.Second*20, time.Minute*1, func() (bool, error) {
```
horis233, line 170: @DanielXLee I am considering if waiting for one minute is too long. How about `PollImmediate(time.Second*2, time.Second*20,`?

controllers/reconcile_operator.go
```diff
@@ -198,27 +198,30 @@ func (r *OperandRequestReconciler) deleteSubscription(operandName string, reques
 		return err
 	}
 
-	if csv != nil {
-		klog.V(2).Infof("Deleting all the Custom Resources for CSV, Namespace: %s, Name: %s", csv.Namespace, csv.Name)
-		if err := r.deleteAllCustomResource(csv, configInstance, operandName, op.Namespace); err != nil {
-			klog.Error("failed to delete a Custom Resource: ", err)
-			return err
-		}
+	if csv == nil {
```
horis233, line 201: Could we just skip the CSV deleting when the CSV is not found?

For example, in this upgrade pending case https://github.com/operator-framework/operator-lifecycle-manager/issues/1548, the CSV will never be created, which will cause these operators can't be removed.

horis233:
@DanielXLee In order to keep the install steps table, could we create another function to fetch the CSV for the deleting as our original plan?

DanielXLee:
👌，I'll split fetch csv in this release.

horis233 (approved):
/lgtm

PR #524, opened 2020-10-08
state the namespace in OperandRegistry

### LINKED ISSUE #520: OperandRequest documentation for namespace is not working as described
/kind bug

Install ODLM at cluster scope on OCP 4.4 (ROKS).  Created an OperandRegistry with a `namespace` of `openshift-operators`, and `scope` of `cluster`.  Created an OperandConfig and OperandRequest.  The requested operator was deployed in the `openshift-operators` namespace, but the operand was also deployed in the `openshift-operators` namespace.

Based on the documentation, I expected the operand to be deployed in the namespace where the OperandRequest was created.

The documentation here: https://github.com/IBM/operand-deployment-lifecycle-manager/blob/master/docs/design/operand-deployment-lifecycle-manager.md#operandregistry-spec
in `[4]` says that the `namespace` is the namespace where the **operator** will be deployed.

The API documentation here:
https://github.com/IBM/operand-deployment-lifecycle-manager/blob/fe9cd5fab224cdfa6061c537e6fe160020ef99b1/api/v1alpha1/operandregistry_types.go#L41
says that the `namespace` is where the operator's **operand** will be deployed.

It appears that the latter is correct, and the former is incorrect, based on my experience.  However I don't know what was intended so I'm hoping that someone can help me sort out the correct behavior, and fix whichever documentation is incorrect.

**Environment:**
- ODLM version: 1.2.3
- Minikube/KIND/OCP version: OCP 4.4.20
- Kubernetes version: (use `kubectl version`): v1.17.1+6af3663
- OS (e.g. from `/etc/os-release`): ?  I don't have access to the nodes, it's on ROKS

Daniel-Fan:
Update docs for resolving the confusion of namespace in OperandRegistry.

Fixes #520 

No

**Release note**:

```release-note

```

### DIFF
--- a/docs/design/operand-deployment-lifecycle-manager.md
+++ b/docs/design/operand-deployment-lifecycle-manager.md
@@ -71,13 +71,13 @@ The Operand (Deployment) Registry Custom Resource (CR) lists OLM Operator inform
 1. `name` of the OperandRegistry
 2. `namespace` of the OperandRegistry
 3. `name` is the name of the operator, which should be the same as the services name in the OperandConfig and OperandRequest.
-4. `namespace` is the namespace where the operator will be deployed.
+4. `namespace` is the namespace where the operator CR will be deployed when InstallMode is `cluster`. Also is the namespace where the operator and operator CR will be deployed when InstallMode is empty or set to `namespace`.
 5. `channel` is the name of OLM channel that is subscribed for the operator.
 6. `packageName` is the name of the package in CatalogSource that is subscribed for the operator.
 7. `scope` is an indicator, either public or private, that dictates whether deployment can be requested from other namespaces (public) or only from the containing names (private). The default is private.
 8. `sourceName` is the name of the CatalogSource.
 9. `sourceNamespace` is the namespace of the CatalogSource.
-10. `installMode` is the install mode of the operator, can be either `namespace` (OLM one namespace) or `cluster` (OLM all namespaces), by default it is `namespace`.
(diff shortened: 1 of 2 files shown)


Daniel-Fan:
@horis233 Could please review again? Thank you!

docs/design/operand-deployment-lifecycle-manager.md
```diff
@@ -71,13 +71,13 @@ The Operand (Deployment) Registry Custom Resource (CR) lists OLM Operator inform
 1. `name` of the OperandRegistry
 2. `namespace` of the OperandRegistry
 3. `name` is the name of the operator, which should be the same as the services name in the OperandConfig and OperandRequest.
-4. `namespace` is the namespace where the operator will be deployed.
+4. `namespace` is the namespace where the operator CR will be deployed when InstallMode is `cluster`. Also is the namespace where the operator and operator CR will be deployed when InstallMode is empty or set to `namespace`.
```
kaczyns, line 74: Could we say "operator's operand" here instead of "operator CR"?  I think given the focus on operands in ODLM it would make more sense to use the term operand here.  It's just a suggestion - feel free to ignore this if you disagree.

horis233: Actually, I discussed this with @Daniel-Fan offline about this yesterday. My only concern for using "operator's operand" is that ODLM creates the CR for an individual operator, but it doesn't know what is the logic of that operator. Even if the CR for an operator is created in namespace A, its operands could be deployed into other namespaces base on the logic from the operator. Thus we can't say operand is deployed in namespace A.

horis233 (approved):
/lgtm

PR #526, opened 2020-10-14
disable leader election to avoid pod restarting

Daniel-Fan:
Remove flag of `enable-leader-election` from command args in order to disable this function.

@chenzhiwei @DanielXLee @horis233 Please correct me if it is wrong.


horis233 (approved):
/lgtm

PR #536, opened 2020-10-27
remove catalogsource check

### LINKED ISSUE #531: Remove ClusterPermission: Check CatalogSource health in openshift-marketplace
/kind feature

We added this check for enhancing the user experience, when CatalogSource is unhealthy, we can output related message to notify users.

How to avoid using cluster permission:

If there is no permission for this checking, the ODLM should skip the checking and throwing messages.

From my point of view, we can add the following logic for this check: 

1. If the ODLM is deployed in the cluster mode, then keep checking (deployed in `openshift-operators` namespace).
2. If the ODLM is deployed in the namespace of the CatalogSource, then keep checking.
3. Otherwise, skip the checking.

^^^ @chenzhiwei @DanielXLee Please review if it is feasible or we have to get the current permission of the ODLM.

horis233:

Remove CatalogSource check to avoid cluster permission.

Fixes #531 

**Release note**:

```release-note

```

### DIFF
--- a/controllers/operandregistry_controller.go
+++ b/controllers/operandregistry_controller.go
@@ -21,9 +21,6 @@ import (
 	"reflect"
 	"time"
 
-	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


controllers/operandregistry_controller.go
```diff
@@ -72,37 +69,19 @@ func (r *OperandRegistryReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
...
-		if err := r.updateOperandRegistryStatus(instance); err != nil {
-			klog.Errorf("failed to update the status for OperandRegistry %s/%s : %v", req.NamespacedName.Namespace, req.NamespacedName.Name, err)
-			return ctrl.Result{}, err
-		}
-		// When catalog source not ready, reconcile every 1 minute to check the status
-		klog.V(2).Info("Waiting for all catalog source ready ...")
-		return ctrl.Result{RequeueAfter: 1 * time.Minute}, nil
+		instance.UpdateRegistryPhase(operatorv1alpha1.RegistryRunning)
+	}
+	if err := r.updateOperandRegistryStatus(instance); err != nil {
```
DanielXLee, line 78: In this method `updateOperandRegistryStatus`,  we use the method `wait.PollImmediate` to update status, it will re-fetch operandregistry, before change in method `updateRegistryOperatorsStatus` will be lost, you can combine following logic in one method
```
	if err := r.updateRegistryOperatorsStatus(instance); err != nil {
		klog.Errorf("failed to update the status for OperandRegistry %s/%s : %v", req.NamespacedName.Namespace, req.NamespacedName.Name, err)

		return ctrl.Result{}, err
	}

	if instance.Status.OperatorsStatus == nil || len(instance.Status.OperatorsStatus) == 0 {
		instance.UpdateRegistryPhase(operatorv1alpha1.RegistryReady)
	} else {
		instance.UpdateRegistryPhase(operatorv1alpha1.RegistryRunning)
	}
	if err := r.updateOperandRegistryStatus(instance); err != nil {
		klog.Errorf("failed to update the status for OperandRegistry %s/%s : %v", req.NamespacedName.Namespace, req.NamespacedName.Name, err)
		return ctrl.Result{}, err
	}
```

horis233: I see. Since our registry logic become simpler, could we just drop the `wait.PollImmediate` method and using the `reconcile.Result` to manage the requeue.

horis233: I believe the change wouldn't be lost. The changed instance is inputted by parameter `newRegistryInstance`. During every time's retry, the function doesn't update the `newRegistryInstance`, but deep copy its status and copy to the current instance.

DanielXLee: you correct.

PR #537, opened 2020-10-29
Shrink permissions for ODLM

horis233:

1. Shrink permissions for ODLM to make it can be deployed in one namespace with few cluster permission.
2. If “INSTALL_SCOPE” is “namespaced”, then create the special cache for ODLM: It will list and watch ODLM CRs from the cluster but only list and watch resource for the namespace of the ODLM operator.
3. Else, it will use a filtered cache to list and watch resource from the cluster scope with a filter to reduce the memory usage.
 
Fixes #

**Release note**:

```release-note

```


DanielXLee (approved):
/lgtm

PR #543, opened 2020-11-02
Update the logic of namespace scope CR management

horis233:

If we delete the namespace from the list when the Operandrequest is deleted, when there are multiple operandrequests in the same namespace, it will cause that when one operandrequest is deleted, its namespace will be removed accidentally. 

Update ns scope management
Fixes #

**Release note**:

```release-note

```


DanielXLee (approved):
/lgtm

PR #559, opened 2020-11-12
remove crds from .gitignore

Daniel-Fan:
Remove `crds` from `.gitignore` file to avoid travis test failed in case bundle.


chenzhiwei (approved):
/lgtm

PR #583, opened 2020-12-29
Use value as the operand Spec

horis233:

This is for avoiding the `null` spec in the operandrequest

Screen Shot 2020-12-28 at 10 50 31 PM

Fixes #

**Release note**:

```release-note

```


api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -129,7 +129,11 @@ func (in *Operand) DeepCopyInto(out *Operand) {
 			(*out)[key] = val
 		}
 	}
-	in.Spec.DeepCopyInto(&out.Spec)
+	if in.Spec != nil {
+		in, out := &in.Spec, &out.Spec
+		*out = new(runtime.RawExtension)
+		(*in).DeepCopyInto(*out)
+	}
```
Daniel-Fan, line 136: May I ask that could we use func (in *RawExtension) DeepCopy() *RawExtension{} directly?

https://github.com/kubernetes/apimachinery/blob/15c5dba13c59f42b25aba4e3a6b17061af770216/pkg/runtime/zz_generated.deepcopy.go#L38

horis233: This file is generated by `controller-gen`, which can't be overwritten by developers.

Daniel-Fan: Oh! I see.

Daniel-Fan (approved):
/lgtm

PR #595, opened 2021-01-18
switch 1.5 to v3 channel

Daniel-Fan:
1. Remove the `replace` section as it is the origin of v3 channel.

2. Keep `olm.skipRange: `'>=1.2.0 <1.5.0'` as it supports upgrading from the version smaller than `1.5.0`


horis233 (approved):
/lgtm

PR #604, opened 2021-02-03
feat: operandregistry status update

horis233:
1. Add labels by patch
2. Add a function to collect all the OperandRequest using a sepcific OperandRegistry

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/controllers/operator/manager.go
+++ b/controllers/operator/manager.go
@@ -120,6 +120,48 @@ func (m *ODLMOperator) ListOperandRequests(ctx context.Context, label map[string
 	return requestList, nil
 }
 
+// ListOperandRequestsByRegistry list all the OperandRequests
+// using the specific OperandRegistry
+func (m *ODLMOperator) ListOperandRequestsByRegistry(ctx context.Context, key types.NamespacedName) (requestList []apiv1alpha1.OperandRequest, err error) {
+	requestCandidates := &apiv1alpha1.OperandRequestList{}
+	if err = m.Client.List(ctx, requestCandidates); err != nil {
+		return
+	}
+	// Set default value for all the OperandRequest
+	for _, item := range requestCandidates.Items {
+		for _, r := range item.Spec.Requests {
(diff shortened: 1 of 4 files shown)


controllers/operator/manager.go
```diff
@@ -120,6 +120,27 @@ func (m *ODLMOperator) ListOperandRequests(ctx context.Context, label map[string
 	return requestList, nil
 }
 
+// ListOperandRequestsForRegistry list all the OperandRequests
+// using the specific operandregistry
+func (m *ODLMOperator) ListOperandRequestsForRegistry(ctx context.Context, registry types.NamespacedName) (requestList []apiv1alpha1.OperandRequest, err error) {
```
bitscuit, line 125: any particular reason for using named return values?

horis233: no. Will it cause an issue?

bitscuit: no functional issues, but format is inconsistent with code base. I think it's a bit less readable

horis233: I believe It is more clear about what is returned and We don't need to initialize return variables.

horis233: @bitscuit Would you mind reviewing it again? cc @Daniel-Fan 
I added the same function for the operandconfig as well.

Daniel-Fan: I would believe it is acceptable to use return value since there are multiple `return` in the function and the return value is not simply `bool` or `error`.

Daniel-Fan (approved):
@bitscuit
if it is OK for you, please add the `lgtm` label

controllers/operator/manager.go
bitscuit: the function is probably short enough that this is okay. We should be careful with using named return values though. It can be hard to figure out what is being returned since the return statement could be far away from where the variable is assigned a value

bitscuit (approved):
/lgtm

PR #608, opened 2021-02-07
feat: Support `protected` prefix in bindinfo.

horis233:
If the key of the bindings map is prefixed with protected, it means the secret and/or configmap can only be shared if it is explicitly declared in the OperandRequest.

Fixes #

**Release note**:

```release-note

```


bitscuit (approved):
/lgtm

PR #610, opened 2021-02-08
handle that not found operator in operandregistry

Daniel-Fan:
should check whether the operator is found before we get its namespace.

When pick the change in `release-1.4` after this one is merged.


horis233 (approved):
/lgtm

Daniel-Fan:
cherry-pick release-1.4

PR #633, opened 2021-03-16
docs: Update documents to make them work with the latest version

horis233:

Fixes #

**Release note**:

```release-note

```


docs/dev/development.md
```diff
@@ -33,26 +33,17 @@
...
+make code-dev
 ```
 
 - Build and push the docker image for local development.
 
 ```bash
-export IMG=<YOUR_CUSTOMIZED_IMAGE_NAME>
-export REGISTRY=<YOUR_CUSTOMIZED_IMAGE_REGISTRY>
-make build-push-images
+export DEV_REGISTRY=<YOUR_CUSTOMIZED_IMAGE_REGISTRY>
```
bitscuit, line 45: Do we need another registry variable? I think Makefile should only use one registry variable if possible.

bitscuit (changes requested):

docs/install/install-with-ocp-3.11.md
```diff
@@ -8,7 +8,10 @@
   - [2. Create CatalogSource](#2-create-catalogsource)
   - [3. Create Operator NS, Group, Subscription](#3-create-operator-ns-group-subscription)
   - [4. Check Operator CSV](#4-check-operator-csv)
-  - [5. Create OperandRequest instance](#5-create-operandrequest-instance)
+  - [5. Create OperandRegistry and OperandConfig instance](#5-create-operandregistry-and-operandconfig-instance)
```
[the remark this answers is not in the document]
horis233: It depends on if OCP3.11 is out of support.

Makefile
```diff
@@ -220,6 +220,12 @@ build-operator-image: ## Build the operator image.
 	--build-arg VCS_REF=$(VCS_REF) --build-arg VCS_URL=$(VCS_URL) \
 	--build-arg GOARCH=$(LOCAL_ARCH) -f Dockerfile .
 
+build-operator-dev-image: ## Build the operator dev image.
```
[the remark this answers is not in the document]
horis233: I think it would be more flexible to have `build-operator-dev-image`, when users/developers what to retag the image after building images.

docs/dev/development.md
horis233: Yes, I think we need another one. Because the other two registries are used for quay.io and artifactory specifically.

bitscuit: We probably should remove the quay one anyways since there is no reason for us to be manually pushing to quay.io/opencloudio. And PROW should always be using artifactory. Developers can just overwrite the artifactory registry variable when testing.

docs/install/install-with-ocp-3.11.md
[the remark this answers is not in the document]
bitscuit: The supported OCP versions should be 4.5 and 4.6, so we shouldn't need to worry about OCP 3.11.

Makefile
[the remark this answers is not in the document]
bitscuit: they would still be able to if combined. Just need to overwrite the variables, e.g.
```
make build-operator-dev-image DEV_REGISTRY=example-registry OPERATOR_IMAGE_NAME=example-name VERSION=dev
```
Would be even easier to retag this way since it's just built with specified tag.

docs/install/install-with-ocp-3.11.md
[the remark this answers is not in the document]
horis233: https://access.redhat.com/support/policy/updates/openshift_noncurrent

OCP 3.11 is supported until June 2020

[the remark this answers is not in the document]
bitscuit: Right, but ODLM does not need to support OCP 3.11 https://www.ibm.com/support/knowledgecenter/SSHKN6/installer/3.x.x/support_os.html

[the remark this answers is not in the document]
horis233: Yes, common service as a product, it doesn't support OCP 3.11. But ODLM as an opensource project it can support vanilla k8s and ocp3.11

docs/dev/development.md
horis233: Why don't we create a specific registry parameter for dev? I believe we need to prevent users from overwriting the artifactory registry which is used for CICD. 

We are using the quay.io for the e2e testing.

Makefile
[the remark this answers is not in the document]
horis233: I don't see the reason why we should combine the command. There is no difference for the end-users and separating can keep the flexiblity.

docs/dev/development.md
bitscuit: The e2e testing should be changed then because it should not be using quay.io/opencloudio. That registry should be reserved for CICD to promote images to since accidental pushes to a release tag would cause problems.

Developers typically would not accidentally push to the integration Artifactory repo because of this check https://github.com/IBM/operand-deployment-lifecycle-manager/blob/master/Makefile#L67-L71. And even if they do, only users with permissions would be able to push there. Pushing to scratch repo is fine since that repo is meant for testing.

docs/install/install-with-ocp-3.11.md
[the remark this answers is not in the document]
bitscuit: Got it

Makefile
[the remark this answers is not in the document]
bitscuit: It's a small quality of life thing. Most development workflow would look like:
```
make build-operator-dev-image
make build-push-dev-image
```
There would not be many scenarios where re-tagging the image is necessary since it's just the developer testing one thing. So if we combine the commands, developer would only need to run one command (for most scenarios):
```
make build-operator-dev-image
```
And the command would just be
```
	@docker build -t $(REGISTRY)/$(OPERATOR_IMAGE_NAME):$(VERSION)\
	--build-arg VCS_REF=$(VCS_REF) --build-arg VCS_URL=$(VCS_URL) \
	--build-arg GOARCH=$(LOCAL_ARCH) -f Dockerfile .
	@docker push $(REGISTRY)/$(OPERATOR_IMAGE_NAME):$(VERSION)
```

If developer needs to re-tag the image to something else, then they could still run the combined command:
```
make build-operator-dev-image REGISTRY=my-registry OPERATOR_IMAGE_NAME=new-name VERSION=new-version
```

If the command was not combined, they would have to run:
```
make build-operator-dev-image
docker tag operator:tag registry/operator:tag
docker push registry/operator:tag
```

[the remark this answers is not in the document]
horis233: Currently, the developer just need to run one command
```
make build-push-dev-image
```

Please check line 237
```
build-push-dev-image: build-operator-dev-image  ## Build and push the operator dev images.
```

docs/dev/development.md
horis233: The e2e testing is planned to be added to the CICD pipeline. Users won't have permission to push any images to the quay.io. 

I agree we should not push e2e test images, build a local image should be fine

horis233: If you still have concern about the quay.io/opencloudio, we can open another quay.io registry for ODLM and installer components

horis233: I agree we shouldn't push e2e images. I will update it in the following commit.

Makefile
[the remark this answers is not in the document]
bitscuit: my bad, I missed that

docs/dev/development.md
bitscuit: Is there any reason the developer cannot just run
```
export ARTIFACTORYA_REGISTRY=<YOUR_CUSTOMIZED_IMAGE_REGISTRY>
```

horis233: Yes, we can, but I believe `ARTIFACTORYA_REGISTRY` will confuse the developer. I believe `DEV_REGISTRY` is a parameter name that can explain itself.

bitscuit: Yes, I agree ARTIFACTORYA_REGISTRY would be confusing, which is why I was wondering whether we can just change the name to `REGISTRY`

horis233:
@bitscuit @Daniel-Fan Any other things we need to update?

docs/dev/e2e.md
```diff
@@ -3,18 +3,18 @@
...
 # Running E2E Tests
 
 1. Ensure **operator-sdk** is installed and login to your OpenShift cluster as an admin user.
 
-1. Run the test using `make test-e2e`  command locally.
+1. Run the test using `make e2e-test-kind`  command locally.
 
     ```bash
-    make test-e2e
+    make e2e-test-kind
```
Daniel-Fan, line 17: I only get one question.

From my understand, does this mean that we need to use `quay.io/opencloud` because the e2e test is planning to integrate into CICD pipeline. 

But this doc and `make e2e-test-kind` is for both developer and CICD, right? Because developer could run `e2e` test and we have removed push command.

Makefile
```diff
@@ -220,6 +219,12 @@ build-operator-image: ## Build the operator image.
...
+	@echo "Building the $(OPERATOR_IMAGE_NAME) docker image..."
+	@docker build -t $(OPERATOR_IMAGE_NAME):$(VERSION) \
+	--build-arg VCS_REF=$(VCS_REF) --build-arg VCS_URL=$(VCS_URL) \
+	--build-arg GOARCH=$(LOCAL_ARCH) -f Dockerfile .
+
 build-test-operator-image: ## Build the operator test image.
 	@echo "Building the $(OPERATOR_IMAGE_NAME) docker image for testing..."
 	@docker build -t $(QUAY_REGISTRY)/$(OPERATOR_IMAGE_NAME):$(OPERATOR_TEST_TAG) \
```
Daniel-Fan, line 230: Also `build-test-operator-image` is used for `e2e-test-kind`, and the image is using `OPERATOR_TEST_TAG` which is `dev-test` in line 83. Could we change the tag to `e2e-test` to distinguish it from `make build-operator-dev-image`?

docs/dev/e2e.md
bitscuit: even if e2e test is integrated into CICD, we wouldn't need quay.io/opencloudio. CICD would just use images from Artifactory

docs/dev/development.md
bitscuit: @horis233 ^

docs/dev/e2e.md
Daniel-Fan: OK, overall I think it is ok and convenient for us to use `quay.io/opencloud` for `e2e` test since it has the non-release tag and we will not push it into the `quay.io/opencloud`

horis233: e2e test won't pull images from the remote image registry. It will just use the locally build image

Makefile
horis233: I believe we should separate them. `build-operator-dev-image` builds the image for developers. `build-test-operator-image` builds the image for CICD.

It keeps flexibility for future change.

docs/dev/development.md
horis233: I think we should separate `DEV_REGISTRY` and `ARTIFACTORYA_REGISTRY`,  when developers export a common variable `registry` it will impact the other make commands. I don't see inconvenience caused by using `DEV_REGISTRY`. Did I miss something?

horis233: @bitscuit If you have an idea about refactoring the Makefile, I am fine with it. 
But this PR focuses on making other opensource contributors try ODLM easily and update our documents.

bitscuit: Yea, that makes sense. This PR should focus on one thing. I think we should try to keep variables to a minimum if possible, otherwise the Makefile gets more complex

Makefile
Daniel-Fan: Sure, we need to keep them separately.

Daniel-Fan (approved):
I don't have any questions about it. Will go to merge it.

/lgtm

PR #634, opened 2021-03-16
Generalize the dependency examples.

cdjohnson:
Generalizing the examples to not have specific operators to avoid confusion.


horis233 (approved):
/lgtm

PR #639, opened 2021-03-21
Support multiOpeReg with same Operator

### LINKED ISSUE #635: handle the same operator existing in multiple OperandRegistry
/kind feature

the two OperandRequests including same lite operator were created from two different OperandRegistry(one is from DV and the other one is from WSL), in this case, once the DV is uninstalled,  the lite operator operandrequest created from DV will be removed. and lite operator will also be uninstalled. But we still have WSL operator which depends on lite operator.

When there is more than one registry which contains the same operator, the ODLM need to handle the uninstallation of that operator well.

Firstly, we need to think about the installation. How to recognize the operator has been installed from one registry, once the operator is requested by another OperandRequest with different registry. It will not duplicate the installation. Currently the ODLM could handle this case.

Secondly, how do we uninstall the operator when there are more than one OperandRequest with different registry containing the same operator? We need avoid ODLM delete the operator if there is at least one OperandRequest exists.
And How do we consider the uninstallation of Operator's CRs under this situation.

Daniel-Fan:

Fixes #635 

TODO:
- [x] Unit Test
- [x] e2e test


horis233:
> @Daniel-Fan: The following test **failed**, say `/retest` to rerun them all:
> 
> Test name	Commit	Details	Rerun command
> kind-e2e-test-operand-deployment-lifecycle-manager-amd64	[b7eb385](https://github.com/IBM/operand-deployment-lifecycle-manager/commit/b7eb385c4a410b855cc7ff29168cbf0056aed657)	[link](http://prow.purple-chesterfield.com/log?job=kind-e2e-test-operand-deployment-lifecycle-manager-amd64&id=1373838986258157569)	`/test e2e-test-kind-operand-deployment-lifecycle-manager-amd64`

The issue could be solved after PR https://github.com/IBM/operand-deployment-lifecycle-manager/pull/640 is merged and rebase on that PR>

controllers/operandrequest/operandrequest_controller_test.go
```diff
@@ -162,17 +182,31 @@ var _ = Describe("OperandRegistry controller", func() {
...
-			Expect(k8sClient.Delete(ctx, request)).Should(Succeed())
+			By("Deleting the first OperandRequest")
+			Expect(k8sClient.Delete(ctx, request1)).Should(Succeed())
+
+			By("Checking jenkins operator have not been deleted")
+			Eventually(func() bool {
+				jenkinsCSV := &olmv1alpha1.ClusterServiceVersion{}
+				err := k8sClient.Get(ctx, types.NamespacedName{Name: "jenkins-csv.v0.0.1", Namespace: operatorNamespaceName}, jenkinsCSV)
+				return err == nil
+			}, testutil.Timeout, testutil.Interval).Should(BeFalse())
```
Daniel-Fan, line 206: For this unit test, I tried to check whether jenkins operator is not deleted, the correct behavior should be `true`, and both `Should(BeFalse())` and `Should(BeTrue())` pass the test.

@horis233

controllers/operandrequest/operandrequest_controller_test.go
```diff
@@ -162,17 +182,31 @@ var _ = Describe("OperandRegistry controller", func() {
...
-			Expect(k8sClient.Delete(ctx, request)).Should(Succeed())
+			By("Deleting the first OperandRequest")
+			Expect(k8sClient.Delete(ctx, request1)).Should(Succeed())
+
+			By("Checking jenkins operator has not been deleted")
+			Eventually(func() error {
+				jenkinsCSV := &olmv1alpha1.ClusterServiceVersion{}
+				err := k8sClient.Get(ctx, types.NamespacedName{Name: "jenkins-csv.v0.0.1", Namespace: operatorNamespaceName}, jenkinsCSV)
+				return err
+			}, testutil.Timeout, testutil.Interval).Should(BeNil())
```
Daniel-Fan, line 206: I created two `OperandRequest` in unit test and both of them have operands `etcd` and `jenkins`, but from different registry.
So if I delete one of the `OperandRequest`. the `etcd` and `jenkins` operator should remain in the cluster. So the `err` returned should `nil` I believe.

It is confusing to me that when I try `return err==nil`, Should(BeFalse())` still passes the test.

And at line 216, after the last `OperandRequest` is deleted, `err` should `isNotFound`.

controllers/operandrequest/operandrequest_controller_test.go
Daniel-Fan: hmmm....Could you take a look at the comment below, thank you! @horis233 
https://github.com/IBM/operand-deployment-lifecycle-manager/pull/639#discussion_r599193928

controllers/operandrequest/operandrequest_controller_test.go
horis233: Since it will retry several times, is it possible in the first the CSV exist but after waiting for a few, it is delered

controllers/operandrequest/operandrequest_controller_test.go
```diff
@@ -160,27 +180,92 @@ var _ = Describe("OperandRegistry controller", func() {
...
+				err := k8sClient.Get(ctx, types.NamespacedName{Name: "jenkins-csv.v0.0.1", Namespace: operatorNamespaceName}, jenkinsCSV)
+				return err == nil
+			}, testutil.Timeout, testutil.Interval).Should(BeFalse())
+
+			By("Checking jenkins operator has been deleted")
+			Eventually(func() bool {
+				jenkinsCSV := &olmv1alpha1.ClusterServiceVersion{}
+				err := k8sClient.Get(ctx, types.NamespacedName{Name: "jenkins-csv.v0.0.1", Namespace: operatorNamespaceName}, jenkinsCSV)
+				return err != nil && errors.IsNotFound(err)
+			}, testutil.Timeout, testutil.Interval).Should(BeTrue())
```
Daniel-Fan, line 262: I used first three ways to check that `jenkins` operator is not deleted, it all pass the test.
and the last two are checking that  `jenkins` operator is deleted, it all pass the test as well.

So the only explanation I could image is that in the first the CSV exists but after waiting for a few, it is deleted.

Here is the passed test result from the last commit
http://prow.purple-chesterfield.com/log?job=test-operand-deployment-lifecycle-manager-amd64&id=1374220756690407425

controllers/operandrequest/operandrequest_controller_test.go
```diff
@@ -160,27 +180,92 @@ var _ = Describe("OperandRegistry controller", func() {
...
+				err := k8sClient.Get(ctx, types.NamespacedName{Name: "etcd-csv.v0.0.1", Namespace: operatorNamespaceName}, etcdCSV)
+				return err
+			}, testutil.Timeout, testutil.Interval).Should(Succeed())
+
+			By("Checking the etcd operator has been deleted")
+			Eventually(func() bool {
+				etcdCSV := &olmv1alpha1.ClusterServiceVersion{}
+				err := k8sClient.Get(ctx, types.NamespacedName{Name: "etcd-csv.v0.0.1", Namespace: operatorNamespaceName}, etcdCSV)
 				return err != nil && errors.IsNotFound(err)
 			}, testutil.Timeout, testutil.Interval).Should(BeTrue())
```
Daniel-Fan, line 206: I did the same test for `etcd`, this time, `etcd` operator is not deleted.
and the second test is failed because it should return `false` but I set it to be `Should(BeTrue())`

So I would assume it is related to `jenkins` operator itself.

Here is the failed test for the latest test
http://prow.purple-chesterfield.com/log?job=test-operand-deployment-lifecycle-manager-amd64&id=1374222266472075265

Daniel-Fan:
@horis233 
Please take a look at the above two comments when you are available, really appreciated it!

controllers/operandrequest/operandrequest_controller_test.go
```diff
@@ -160,27 +180,78 @@ var _ = Describe("OperandRegistry controller", func() {
...
-			By("Deleting the OperandRequest")
-			Expect(k8sClient.Delete(ctx, request)).Should(Succeed())
-
-			By("Checking operators have been deleted")
+			By("Checking etcd operator has been deleted")
 			Eventually(func() bool {
 				etcdCSV := &olmv1alpha1.ClusterServiceVersion{}
 				err := k8sClient.Get(ctx, types.NamespacedName{Name: "etcd-csv.v0.0.1", Namespace: operatorNamespaceName}, etcdCSV)
 				return err != nil && errors.IsNotFound(err)
 			}, testutil.Timeout, testutil.Interval).Should(BeTrue())
```
Daniel-Fan, line 248: If I did not disable `etcd` operator before, so both `OperandRequest` have `etcd` and `jejenkins`.

Then directly start to delete the first `OperandRequest`, this time `jenkins` operator exists and line 241 failed because it should be `Should(BeFalse())`

So it may not the problem of `jenkins` operator.

Here is the failed test link
http://prow.purple-chesterfield.com/log?job=test-operand-deployment-lifecycle-manager-amd64&id=1374232080648507394

Daniel-Fan:
Anyway, I still add the unit test for supporting multi registry. following the correct logic, now It could pass the test on prow.

Also I manually go through the test procedure on the cluster, it has the normal behavior.

But it is still weird that sometimes `jenkins` operator could pass both positive and negative tests about checking its existence.

controllers/operandrequest/reconcile_operator.go
```diff
@@ -185,11 +185,21 @@ func (r *Reconciler) createSubscription(ctx context.Context, cr *operatorv1alpha
 	return nil
 }
 
-func (r *Reconciler) updateSubscription(ctx context.Context, cr *operatorv1alpha1.OperandRequest, sub *olmv1alpha1.Subscription) error {
+func (r *Reconciler) updateSubscription(ctx context.Context, cr *operatorv1alpha1.OperandRequest, sub *olmv1alpha1.Subscription, key types.NamespacedName) error {
 
 	klog.V(2).Infof("Updating Subscription %s/%s ...", sub.Namespace, sub.Name)
 	cr.SetUpdatingCondition(sub.Name, operatorv1alpha1.ResourceTypeSub, corev1.ConditionTrue)
 
+	sub.Annotations[key.Namespace+"."+key.Name] = "true"
```
horis233, line 193: @Daniel-Fan I believe we need to add a check to make sure if `sub.Annotations` exists to make it work for the upgrade situation.

controllers/operandrequest/reconcile_operator.go
```diff
@@ -185,11 +185,21 @@ func (r *Reconciler) createSubscription(ctx context.Context, cr *operatorv1alpha
...
+	mergePatch, _ := json.Marshal(map[string]interface{}{
+		"metadata": map[string]interface{}{
+			"annotations": sub.Annotations,
+		},
+	})
+	if err := r.Patch(ctx, sub, client.RawPatch(types.MergePatchType, mergePatch)); err != nil {
```
horis233, line 199: Why we patch it oncee and update it once?

controllers/testutil/test_util.go
```diff
@@ -34,7 +34,7 @@ import (
 
 // CreateNSName generates random namespace names. Namespaces are never deleted in test environment
 func CreateNSName(prefix string) string {
-	suffix := make([]byte, 20)
+	suffix := make([]byte, 10)
```
horis233, line 37: I am fine with this change, but any reason why we do that?

controllers/operandrequest/reconcile_operator.go
Daniel-Fan: hmmm....I did not get the point.

If it is the upgrade scenario, after the ODLM is upgraded, it will reconcile all the resource instance, is that right? And it will add the `annotations` for each subscription based on `registry` in `OperandRequest`.

Is it the same as the fresh installation?

controllers/operandrequest/reconcile_operator.go
Daniel-Fan: My bad, I ignore the subscription update itself.
Yes, we could update annotations at line 108 with updates of other `spec` field.

controllers/testutil/test_util.go
Daniel-Fan: For the annotation `key/value`, the `key` should be less than 63 characters.
So if we use 20 bytes suffix, it will convert to 40 characters because of %x, base 16, two characters per byte.
`key.Namespace+"."+key.Name` will be longer than 63.

controllers/operandrequest/reconcile_operator.go
horis233: Yes, but existing subscriptions don't have `Annotations` field. it will return Invalid memory address or nil pointer dereference issue.

We need to make a map first.

controllers/testutil/test_util.go
horis233: Oh, good point. Nice to know that.

controllers/operandrequest/reconcile_operator.go
Daniel-Fan: Get it!
Tested it in a upgrade scenario. It indeed will cause the panic of nil pointer dereference.
Really thank you to point it out!

controllers/operandrequest/reconcile_operator.go
```diff
@@ -97,14 +97,19 @@ func (r *Reconciler) reconcileOperator(ctx context.Context, requestInstance *ope
...
 						sub.Spec.CatalogSourceNamespace = opt.SourceNamespace
 						sub.Spec.Package = opt.PackageName
 						if opt.InstallPlanApproval != "" && sub.Spec.InstallPlanApproval != opt.InstallPlanApproval {
 							sub.Spec.InstallPlanApproval = opt.InstallPlanApproval
 						}
+						// initialize map interface for upgrade case
+						if sub.Annotations == nil {
+							sub.Annotations = make(map[string]string)
+						}
+						sub.Annotations[registryKey.Namespace+"."+registryKey.Name] = "true"
```
Daniel-Fan, line 112: @horis233 
PR is updated, for the existing subscription, handle the upgrade scenario and update annotations with other `sepc` field in the same place.

Daniel-Fan:
@bitscuit 

PR is ready for merging, could you please review it?

controllers/operandrequest/reconcile_operator.go
```diff
@@ -97,14 +97,19 @@ func (r *Reconciler) reconcileOperator(ctx context.Context, requestInstance *ope
...
-					if compareSub(sub.Spec, opt) {
+					if compareSub(sub, opt, registryKey) {
 						sub.Spec.CatalogSource = opt.SourceName
 						sub.Spec.Channel = opt.Channel
 						sub.Spec.CatalogSourceNamespace = opt.SourceNamespace
 						sub.Spec.Package = opt.PackageName
 						if opt.InstallPlanApproval != "" && sub.Spec.InstallPlanApproval != opt.InstallPlanApproval {
 							sub.Spec.InstallPlanApproval = opt.InstallPlanApproval
 						}
+						// initialize map interface for upgrade case
```
bitscuit, line 108: comment isn't that useful, since I can already see that a map is being initialized below; recommend changing to

```suggestion
						// add annotations to existing Subscriptions for upgrade case
```

controllers/operandrequest/reconcile_operator.go
```diff
@@ -218,6 +223,27 @@ func (r *Reconciler) deleteSubscription(ctx context.Context, operandName string,
 		return nil
 	}
 
+	// check and remove registry in annotation of subscription
+	regName := registryInstance.ObjectMeta.Name
+	regNs := registryInstance.ObjectMeta.Namespace
+	delete(sub.Annotations, regNs+"."+regName)
+	if len(sub.Annotations) != 0 {
+		// remove the registry from subscription and remain the resource present
```
bitscuit, line 231: maybe change comment to be more clear

```suggestion
		// remove the associated registry from annotation of subscription
```

bitscuit (changes requested):

controllers/operandrequest/reconcile_operator.go
Daniel-Fan: Good point! updated.

Daniel-Fan:
@bitscuit 
The comments are updated, please review it again, Thank you!

bitscuit (approved):
/lgtm

PR #645, opened 2021-04-03
feat: Apply the ALM example increment to existing CRs 

### LINKED ISSUE #637: Updating the existing CRs during the upgrade
/kind feature

Currently, during the operator upgrade, ODLM won't update the existing CR according to the changes in the operator `alm-example`.

The existing upgrade logic is:

The updated CR will be generated by the existing CR in the cluster and the configurations in the OperandConfig. The field in the OperandConfig will overwrite the existing CR when it is set in both OperandConfig and the existing CR. We can say OperandConfig has a higher priority than the existing CR.

In new upgrading logic, in order to pick up the new parameter added into the alm-example, ODLM will compare the alm-example, existing CR and the configuration in the OperandConfig.

Among these three files, alm-example has the lowest priority, the priority of the existing CR is higher and the priority of the OperandConfig is the highest. It means only the new field in the alm-example will be picked up to the final CR, other fields would be overwritten by the existing CR or OperandConfig.

horis233:
fix #637

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/controllers/operandbindinfo/operandbindinfo_controller.go
+++ b/controllers/operandbindinfo/operandbindinfo_controller.go
@@ -85,7 +85,7 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reconcileErr er
 		}
 	}()
 
-	klog.V(1).Infof("Reconciling OperandBindInfo: %s", req.NamespacedName)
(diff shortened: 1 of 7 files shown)


horis233:
@bitscuit @Daniel-Fan 
Could you help review this PR?

Daniel-Fan (approved):
It looks good to me. I am gonna approve it.

PR #654, opened 2021-04-13
fetch ODLM_SCOPE and cache ODLM CRs

Daniel-Fan:
We will create an another ConfigMap called `odlm-scope` to manage the scope for ODLM.

It includes `ControlNs` comparing with the previous `namespace-scope` ConfigMap. We will remove `ControlNs` from `common-service` NSS CR.

```
apiVersion: v1
data:
  namespaces: placeholder
kind: ConfigMap
metadata:
  name: odlm-scope
  namespace: placeholder
```

Namespace-scope operator will have an additional CR called `nss-odlm-scope` to update the above ConfigMap, to make sure that scope of ODLM is in the valid namespaces.

```
apiVersion: operator.ibm.com/v1
kind: NamespaceScope
metadata:
  name: nss-odlm-scope
  namespace: placeholder
spec:
  namespaceMembers:
  - placeholder
  configmapName: odlm-scope
  restartLabels:
    intent: projected-odlm
```

Therefore `WATCH_NAMESPACE` in the csv in ODLM will change to watch the namespaces in ConfigMap `odlm-scope`.

Also we only cache the ODLM CRs in the scope, so ODLM will not watch the request out of scope.

### DIFF
--- a/controllers/util/util.go
+++ b/controllers/util/util.go
@@ -50,12 +50,12 @@ func GetInstallScope() string {
 	return ns
 }
 
-func GetOdlmScope() string {
+func GetOdlmScope() bool {
 	isEnable, found := os.LookupEnv("ODLM_SCOPE")
-	if !found {
-		return "false"
+	if !found || isEnable != "true" {
+		return false
 	}
-	return isEnable
+	return true
 }
 
 // ResourceExists returns true if the given resource kind exists
--- a/controllers/util/util.go
+++ b/controllers/util/util.go
@@ -50,6 +50,14 @@ func GetInstallScope() string {
 	return ns
 }
 
+func GetOdlmScope() string {
+	isEnable, found := os.LookupEnv("ODLM_SCOPE")
+	if !found {
+		return "false"
+	}
+	return isEnable
+}
+
 // ResourceExists returns true if the given resource kind exists
 // in the given api groupversion
 func ResourceExists(dc discovery.DiscoveryInterface, apiGroupVersion, kind string) (bool, error) {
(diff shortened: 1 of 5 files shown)


Daniel-Fan:
This PR is ready for reviewing. @horis233 @bitscuit

main.go
```diff
@@ -91,9 +91,17 @@ func main() {
 
 	scope := util.GetInstallScope()
 	watchNamespace := util.GetWatchNamespace()
+	odlmScopeEnable := util.GetOdlmScope()
 	if scope == "namespaced" {
-		options.NewCache = k8sutil.NewODLMCache(strings.Split(watchNamespace, ","), gvkLabelMap)
+		// single ODLM instance
```
bitscuit, line 96: don't really need this comment or should change to "namespaced ODLM"

controllers/util/util.go
```diff
@@ -50,6 +50,14 @@ func GetInstallScope() string {
 	return ns
 }
 
+func GetOdlmScope() string {
```
bitscuit, line 53: any reason why this function doesn't return a boolean instead?

bitscuit (changes requested):

controllers/util/util.go
Daniel-Fan: Great point, updated!

Daniel-Fan:
@horis233 @bitscuit @ZhuoxiLi 
This PR is ready for reviewing.

bitscuit (approved):
/lgtm

PR #655, opened 2021-04-16
feat: ODLM can check the existing subscription by packagenames

### LINKED ISSUE #652: ODLM should be able to find existing operator by packageName
/kind feature

If a subscription A is created in namespace-a via OLM.
And then create an OperandRequest for an Operand that has package A defined, ODLM should be able to detect that and just re-use it.

horis233:

Fixes #652 

**Release note**:

```release-note

```

### DIFF
--- a/controllers/operator/manager.go
+++ b/controllers/operator/manager.go
@@ -18,10 +18,12 @@ package operator
 
 import (
 	"context"
+	"fmt"
 
 	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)


controllers/operator/manager.go
```diff
@@ -191,8 +193,8 @@ func (m *ODLMOperator) ListOperandRequestsByConfig(ctx context.Context, key type
 	return
 }
 
-// GetSubscription gets Subscription from a name
-func (m *ODLMOperator) GetSubscription(ctx context.Context, name, namespace string, packageName ...string) (*olmv1alpha1.Subscription, error) {
+// GetSubscription gets Subscription by name and package name
+func (m *ODLMOperator) GetSubscription(ctx context.Context, name, namespace string, packageName string) (*olmv1alpha1.Subscription, error) {
```
Daniel-Fan, line 197: Could you merge all three `string` parameters to `(name, namespace, packageName string)`?

controllers/operator/manager.go
```diff
@@ -205,12 +207,25 @@ func (m *ODLMOperator) GetSubscription(ctx context.Context, name, namespace stri
...
 	}
-	subPkgKey := types.NamespacedName{
-		Name:      packageName[0],
-		Namespace: namespace,
+
+	subList := &olmv1alpha1.SubscriptionList{}
+	if err := m.Client.List(ctx, subList, &client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{
+			"operators.coreos.com/" + packageName + "." + namespace: "",
+		}),
```
Daniel-Fan, line 215: We change from checking the label in the subscription instead of the subscription's name by `packageName`.

Is it because the subscription created via OLM, will always have those kind of `package` label. But the name of subscription is uncertain, it may vary since it could be decided by user?

controllers/operator/manager.go
horis233: Nice catch

controllers/operator/manager.go
horis233: Yes, the label is added by OLM. I am wondering if we should check these labels or just iterate all the subscriptions in the namespace.

Daniel-Fan: I think the current method is more elegant.

If we iterate the subscriptions in that namespace, we still need to check the labels in them. So why don't we just `List` the subscriptions with those labels and count their numbers.

Daniel-Fan (approved):
It looks good to me. I am gonna approve it.

Daniel-Fan (approved):
/lgtm

PR #659, opened 2021-04-21
add complete err check

Daniel-Fan:
@horis233 @bitscuit 

Does this defect fix need to be added into `release-3.7`?


bitscuit:
> 
> 
> Does this defect fix need to be added into `release-3.7`?

I don't think so. It's not an urgent fix, and 3.7.3 fix pack has been delayed already, so let's avoid making any unnecessary changes

bitscuit (approved):
/lgtm

horis233:
@bitscuit 

How about we cherry-pick the fix to the 3.7 branch and hold it until we released 3.7.3?

bitscuit:
sure, though it's still unknown whether we will have 3.7.4

PR #661, opened 2021-04-22
Reconcile two NSS CRs managedby ODLM

Daniel-Fan:
ODLM will add the namespace which are captured from the `Namespace` where Operator will be installed in `OperandRegistry` and `Namespace` of `OperandRequest` into both `nss-managedby-odlm` and `odlm-scope-managedby-odlm` NSS CRs.

On the other word, those two NSS CRs has the same `namespaceMember`

The first `nss-managedby-odlm` is used to control namespace-scope for individual CS operators.
The latter `odlm-scope-managedby-odlm` is used to control namespace-scope for ODLM.

Also those two NSS CRs will only take effect in on-prem Single CS instance case.

If the `odlmScopeEnable` is `true`, indicating the SaaS or om-prem multi instances case, `namespacescope_controller` will be disabled and above two CRs will not be generated. See details here
https://github.com/IBM/ibm-common-service-operator/pull/490

### DIFF
--- a/controllers/namespacescope/namespacescope_controler_test.go
+++ b/controllers/namespacescope/namespacescope_controler_test.go
@@ -49,6 +49,7 @@ var _ = Describe("NamespaceScope controller", func() {
 		request              *operatorv1alpha1.OperandRequest
 		config               *operatorv1alpha1.OperandConfig
 		nss                  *nssv1.NamespaceScope
+		odlmNss              *nssv1.NamespaceScope
 	)
 
 	BeforeEach(func() {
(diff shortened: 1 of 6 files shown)


controllers/namespacescope/namespacescope_controler_test.go
```diff
@@ -102,6 +114,14 @@ var _ = Describe("NamespaceScope controller", func() {
 				}
 				return len(nss.Spec.NamespaceMembers)
 			}, timeout, interval).Should(Equal(3))
```
Daniel-Fan, line 116: @horis233 

May I ask why we expect `len(NamespaceMembers)==3`?

From my understanding, the controller will add the `namespace` of all `OperandRequest` and `operator.namespace` in `OperandRegistry`.

So in this case, there is one `OperandRequest` and `namespace` of two operators `etcd` and `jenkins` has the same. The total number of namespace should be `2`.

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/operator-framework/operator-lifecycle-manager v0.17.0
 	github.com/pkg/errors v0.9.1
 	k8s.io/api v0.18.9
+	k8s.io/apiextensions-apiserver v0.18.9
```
horis233, line 16: Let's remove this line, it is only used in the test instead of product code.

controllers/namespacescope/namespacescope_controler_test.go
horis233: It's because it added the namespace of the operator as well.
You can use `fmt` to print the namespace list to check.

go.mod
Daniel-Fan: Sure, after deleting `testbin`, `go.mod` does not include this module.

controllers/namespacescope/namespacescope_controler_test.go
Daniel-Fan: Check it, indeed there are three namespaces

horis233 (approved):
/lgtm

PR #669, opened 2021-05-02
feat: Install operators in batch.

### LINKED ISSUE #624: Deploy operator CR in batch
A follow up issue for https://github.com/IBM/operand-deployment-lifecycle-manager/issues/622#issuecomment-790136949

@horis233

horis233:
user can use the flag "-batch-chunk-size" to control how many subscriptions will be deployed simultaneously

Fixes #624 

**Release note**:

```release-note

```

### DIFF
--- a/main.go
+++ b/main.go
@@ -73,6 +73,8 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	var stepSize = flag.Int("batch-chunk-size", 3, "batch-chunk-size is used to control at most how many subscriptions will be created concurrently")
+
 	flag.Parse()
 
 	gvkLabelMap := map[schema.GroupVersionKind]cache.Selector{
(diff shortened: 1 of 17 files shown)


main.go
```diff
@@ -73,6 +73,8 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	var stepSize = flag.Int("batch-chunk-size", 3, "batch-chunk-size is used to control at most how many subscriptions will be created concurrently")
```
Daniel-Fan, line 76: May I ask that if `batch-chunk-size` is the flag, does it mean we need pre-set it in our CSV as well?
https://github.com/IBM/operand-deployment-lifecycle-manager/blob/master/bundle/manifests/operand-deployment-lifecycle-manager.clusterserviceversion.yaml#L283

horis233: It has a default value "3". It is not a necessary flag

Daniel-Fan: Oh I see, thanks!

horis233:
/hold this PR 

1_bZHBo75FSyKre5pk2-HPmw

Too many locks used in this change, we need a better method to manage the batch install.

horis233:
@Daniel-Fan @ZhuoxiLi 

This PR is ready for review.

This PR is changed much from the last review. It still uses locks to update status. I tried pushing status update into a channel, but the effect is not very good.

Let's use the lock as version 1. We can make further refactor in future PRs.

horis233:
Looks, it didn't pass the e2e test. I will make more testings on it

horis233:
@Daniel-Fan @ZhuoxiLi 

It finally passes all the tests. Please review this PR. 

We can take it the stage 1 of the concurrence support. We can do more reactors on it in the future.

Daniel-Fan (approved):
/lgtm

PR #673, opened 2021-05-07
check ODLM priority when reconciling resource

Daniel-Fan:
Although `ibm-common-service-operator` will pass the same resource values for cluster scope resource into the `OperandConfig` of each ODLM, implementing here https://github.com/IBM/ibm-common-service-operator/pull/510
We still allow only one ODLM to reconcile those cluster scope resource in order to reduce the resource consumption.

For both namespace scoped and cluster scoped resource, when a new `OperandRequest` is created for operators and operands. The subscription of operators will be added a annotation about its corresponding `OperandConfig` to indicate that the resources will follow which configuration(s).

**Only the ODLM whose `OperandConfig` was firstly fetched from the subscription's annotation is able to reconcile this resource.**

The annotation will be in the following format
1. For namespace scope resources, only one annotation about `OperandConfig` will exist
   ```yaml
    annotations:
        cp4i-common-services.common-service/config: true
    ```
2. For cluster scope resources, multiple annotations about `OperandConfig` will exist
    ```yaml
    annotations:
        cp4i-common-services.common-service/config: true
        cp4d-common-services.common-service/config: true
        cp4s-common-services.common-service/config: true
    ```
In the above case for cluster scope resources, the ODLM in `cp4i-common-services` namespace has the reconciliation priority of this resource.

#### What to do if the reconciling ODLM does not has OperandRequest about cluster resources anymore(OperandRequest is deleted)
When the `OperandRequest` about cluster scope resources is deleted in the reconciling ODLM, this ODLM should not reconcile this resources anymore.
In order to achieve this, the corresponding annotation about `OperandConfig` will be evicted, one of the ODLM in the remaining annotations about `OperandConfig` will take over the reconciliation.

For example, the `OperandRequest` from `cp4i-common-services` is removed, the updated annotations will look like this
```yaml
annotations:
    cp4d-common-services.common-service/config: true
    cp4s-common-services.common-service/config: true
```
the ODLM in `cp4d-common-services` namespace now has the control of this resource.

### DIFF
--- a/controllers/operandrequest/reconcile_operand.go
+++ b/controllers/operandrequest/reconcile_operand.go
@@ -23,6 +23,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -64,6 +65,9 @@ func (r *Reconciler) reconcileOperand(ctx context.Context, requestInstance *oper
 			merr.Add(errors.Wrapf(err, "failed to get the OperandRegistry %s", registryKey.String()))
 			continue
 		}
+		regName := registryInstance.ObjectMeta.Name
+		regNs := registryInstance.ObjectMeta.Namespace
+
 		for i, operand := range req.Operands {
 
 			opdRegistry := registryInstance.GetOperator(operand.Name)
@@ -93,6 +97,20 @@ func (r *Reconciler) reconcileOperand(ctx context.Context, requestInstance *oper
 			if _, ok := sub.Labels[constant.OpreqLabel]; !ok {
 				// Subscription existing and not managed by OperandRequest controller
 				klog.Warningf("Subscription %s in the namespace %s isn't created by ODLM", sub.Name, sub.Namespace)
+			}
+
+			// check config annotation in subscription, identify the first ODLM has the priority to reconcile
+			var firstMatch string
+			reg, _ := regexp.Compile(`^(.*)\.(.*)\/config`)
+			for anno := range sub.Annotations {
+				if reg.MatchString(anno) {
+					firstMatch = anno
+					break
+				}
+			}
+			if firstMatch != "" && firstMatch != regNs+"."+regName+"/config" {
+				klog.Infof("Subscription %s in the namespace %s is currently managed by %s", sub.Name, sub.Namespace, firstMatch)
+				continue
 			}
 
 			csv, err := r.GetClusterServiceVersion(ctx, sub)
--- a/controllers/operandrequest/reconcile_operator.go
+++ b/controllers/operandrequest/reconcile_operator.go
@@ -113,6 +113,7 @@ func (r *Reconciler) reconcileOperator(ctx context.Context, requestInstance *ope
 						}
 						sub.Annotations[registryKey.Namespace+"."+registryKey.Name+"/registry"] = "true"
 						sub.Annotations[requestInstance.Namespace+"."+requestInstance.Name+"/request"] = "true"
+						sub.Annotations[registryKey.Namespace+"."+registryKey.Name+"/config"] = "true"
 						if err = r.updateSubscription(ctx, requestInstance, sub); err != nil {
 							requestInstance.SetMemberStatus(opt.Name, operatorv1alpha1.OperatorFailed, "")
 							return err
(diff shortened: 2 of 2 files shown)


Daniel-Fan:
Previously I think we need to add annotations for both operator and operand. Later on I noticed we always fetch the corresponding operands through their operator subscription.

Therefore, we only need to add annotations into each subscription which simplify coding complexity a lot.

Please hold those two PRs, need further test for verifying.

Daniel-Fan:
@horis233 @ZhuoxiLi During the test, found a functionality regression

1. Previously we support case that the user to update hardware profile in the `OperandConfig` directly. But this case is not available for cluster scope resource in multi instance anymore, because the `OperandConfig` update for cluster scope resource in one of the namespaced ODLM will not be synced to other scoped `OperandConfig`. If the user want to update the profile, they have to updating the CommonService CR.

Any comments for the above situation? If we still want to sync up the hardware profile between different scoped `OperandConfig` for the cluster scope resource which is changed directly in `OperandConfig`, we have to let namespaced ODLM watch all the CRs through the cluster.


---------------------------------------------The below is not PR related------------------------------------------
Also get a question for annotations in subscription about this PR https://github.com/IBM/operand-deployment-lifecycle-manager/pull/645

There is a request label added for subscription, here is the example after I created two `OperandRequest` and delete one of them.
```
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  annotations:
    cs-installed-1.common-service/request: 'true'
    cs-installed-2.common-service/config: 'true'
    cs-installed-2.common-service/registry: 'true'
    cs-installed-2.common-service/request: 'true'
```
Basically we clean up the registry and config label for deleted `OperandRequest` in `cs-installed-1` namespace.
But `cs-installed-1.common-service/request: 'true'` is still there, any particular reason about it?

If it is too complicated or time-consuming to explain, maybe we could discuss in the sync up meeting.

horis233:
> Previously we support case that the user to update hardware profile in the OperandConfig directly. But this case is not available for cluster scope resource in multi instance anymore, because the OperandConfig update for cluster scope resource in one of the namespaced ODLM will not be synced to other scoped OperandConfig. If the user want to update the profile, they have to updating the CommonService CR.

I think we never suggest users update hardware profiles from the OperandConfig. if they do that, the hardware profile will be overwritten by commonservice CR anyway.



Labels like `cs-installed-1.common-service/request: 'true'` are used to build a mapping from OperandRequest to subscriptions. It is not used to manage the CR/operator deletion, so we don't remove this label.

Daniel-Fan:
> I think we never suggest users update hardware profiles from the OperandConfig. if they do that, the hardware profile will be overwritten by commonservice CR anyway.

That's good if users will not update `OperandConfig` directly, then we would not have to concern about this issue. But from my understanding, if the value in `OperandConfig` is larger than the one in any of the Common Service CRs, it will keep the largest one in `OperandConfig`, is that right?

> Labels like cs-installed-1.common-service/request: 'true' are used to build a mapping from OperandRequest to subscriptions. It is not used to manage the CR/operator deletion, so we don't remove this label.

Oh I see, but if the `OperandRequest` has been removed in this case, should it keep label in subscription? I am concerned that if the next time a new `OperandRequests` is created with the same name but for different operator, I assume that it will cause some issue?

horis233:
> > I think we never suggest users update hardware profiles from the OperandConfig. if they do that, the hardware profile will be overwritten by commonservice CR anyway.
> 
> That's good if users will not update `OperandConfig` directly, then we would not have to concern about this issue. But from my understanding, if the value in `OperandConfig` is larger than the one in any of the Common Service CRs, it will keep the largest one in `OperandConfig`, is that right?
> 


I don't think so. Otherwise, we will never scale the size down.

> > Labels like cs-installed-1.common-service/request: 'true' are used to build a mapping from OperandRequest to subscriptions. It is not used to manage the CR/operator deletion, so we don't remove this label.
> 
> Oh I see, but if the `OperandRequest` has been removed in this case, should it keep label in subscription? I am concerned that if the next time a new `OperandRequests` is created with the same name but for different operator, I assume that it will cause some issue?

Ah, I didn't notice that. It is a good catch. That is a place we can enhance, please create an issue for that.

ZhuoxiLi (approved):
/lgtm

PR #677, opened 2021-05-11
update ocp to Redhat supports versions

ZhuoxiLi:
OCP 4.2 or 4.3 are no longer supported by RadHat, updated ocp version to supported versions

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,14 +40,14 @@ Operand Deployment Lifecycle Manager has four CRDs:
 
 ## Supported platforms
 
-You can install the Operand Deployment Lifecycle Manager on Linux® x86_64 with Red Hat® OpenShift® Container Platform version 4.5+.
+You can install the Operand Deployment Lifecycle Manager on Linux® x86_64 with Red Hat® OpenShift® Container Platform version 4.3+.
 
 ## Prerequisites
 
-- [operator-sdk][operator_sdk] version v0.19.2.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -40,14 +40,14 @@ Operand Deployment Lifecycle Manager has four CRDs:
 
 ## Supported platforms
 
-You can install the Operand Deployment Lifecycle Manager on Linux® x86_64 with Red Hat® OpenShift® Container Platform version 4.2 or 4.3.
+You can install the Operand Deployment Lifecycle Manager on Linux® x86_64 with Red Hat® OpenShift® Container Platform version 4.5+.
```
horis233, line 43: How about changing it to 4.x? Otherwise we need to update it every time OCP version is out of support.

horis233: Probably should be 4.3+, because CRD v1 is supported in 4.3

ZhuoxiLi: sure, maybe change to 4.3+. but.. why CRD v1 still supports OCP versions that are deprecated by Redhat？

horis233: It's because RedHat support lifecycle doesn't come with k8s updates. An OCP version isn't supported by Redhat, which doesn't mean ODLM can work on it. From an open-source software perspective, ODLM can still support that.

horis233 (approved):
/lgtm

PR #678, opened 2021-05-14
fix: Make operandrequest can be removed together with the namespace d…

### LINKED ISSUE #649: OperandRequests with Finalizers block namespace deletion
/kind bug

1. Create a namespace (named `ptest` in my test)
2. Create an OperandRequest in the namespace (for an `licensing` service in my test)
3. Wait for the OperandRequest to result in the creation of a configmap and secret in the namespace (provides service url and token for the service requested)
4. Delete the namespace
5. The namespace is stuck in 'terminating' state because the OperandRequest cannot be deleted.  The OperandRequest cannot be deleted because it still has the `finalizer.request.ibm.com` finalizer.

The Finalizer should be automatically removed from the OperandRequest and the namespace able to be deleted cleanly.

The log from the `operand-deployment-lifecycle-manager-[id]` pod has many messages like this upon deleting the namespace:
```
I0408 22:45:21.631673       1 operandbindinfo_controller.go:88] Reconciling OperandBindInfo: ibm-common-services/ibm-licensing-bindinfo
E0408 22:45:22.048797       1 operandbindinfo_controller.go:200] failed to reconcile the OperandBindinfo ibm-common-services/ibm-licensing-bindinfo: the following errors occurred:
  - failed to create secret ptest/ibm-licensing-bindinfo-ibm-licensing-token: secrets is forbidden: User "system:serviceaccount:ibm-common-services:operand-deployment-lifecycle-manager" cannot create resource "secrets" in API group "" in the namespace "ptest"
  - failed to create secret ptest/dummyenv-metering-lssecret: secrets is forbidden: User "system:serviceaccount:ibm-common-services:operand-deployment-lifecycle-manager" cannot create resource "secrets" in API group "" in the namespace "ptest"
I0408 22:45:22.682684       1 request.go:621] Throttling request took 1.043243216s, request: GET:https://172.30.0.1:443/apis/config.openshift.io/v1?timeout=32s
I0408 22:45:23.060068       1 operandbindinfo_controller.go:88] Reconciling OperandBindInfo: ibm-common-services/ibm-licensing-bindinfo
E0408 22:45:23.385037       1 operandbindinfo_controller.go:200] failed to reconcile the OperandBindinfo ibm-common-services/ibm-licensing-bindinfo: the following errors occurred:
  - failed to create secret ptest/ibm-licensing-bindinfo-ibm-licensing-token: secrets is forbidden: User "system:serviceaccount:ibm-common-services:operand-deployment-lifecycle-manager" cannot create resource "secrets" in API group "" in the namespace "ptest"
  - failed to create secret ptest/dummyenv-metering-lssecret: secrets is forbidden: User "system:serviceaccount:ibm-common-services:operand-deployment-lifecycle-manager" cannot create resource "secrets" in API group "" in the namespace "ptest"
W0408 22:45:24.140279       1 operandrequest_controller.go:98] No permission to update OperandRequest
```
To _my_ eyes it is trying to reconcile the OperandRequest because the Secret was deleted during (attempted) namespace deletion.  But it cannot recreate the secret because the namespace is being deleted.  I _assume_ it needs some logic to check the namespace when reconciling.  If the namespace.metadata.deletionTimestamp is not nil, the namespace is being deleted.  If the namespace is being deleted then instead of trying to create/update the operands, the reconcile loop should just remove the finalizer from the OperandRequest.

**Environment:**

- ODLM version: `operand-deployment-lifecycle-manager.v1.5.0`
- Minikube/KIND/OCP version: `OCP 4.3.40`
- Kubernetes version: (use `kubectl version`): `v1.16.2+853223d`
- OS (e.g. from `/etc/os-release`): n/a

horis233:
…eletion

Grant ODLM extra permission to patch the remaining OperandRequest to remove the finalizer from it.

Fixes #649

**Release note**:

```release-note

```

### DIFF
--- a/controllers/operandrequest/operandrequest_controller.go
+++ b/controllers/operandrequest/operandrequest_controller.go
@@ -84,6 +84,28 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reconcileErr er
 		}
 	}()
 
+	// Remove finalizer when DeletionTimestamp none zero
+	if !requestInstance.ObjectMeta.DeletionTimestamp.IsZero() {
+
+		// Check and clean up the subscriptions
+		err := r.checkFinalizer(ctx, requestInstance)
+		if err != nil {
+			klog.Errorf("failed to clean up the subscriptions for OperandRequest %s: %v", req.NamespacedName.String(), err)
+			return ctrl.Result{}, err
+		}
+
+		originalReq := requestInstance.DeepCopy()
+		// Update finalizer to allow delete CR
+		if requestInstance.RemoveFinalizer() {
+			err = r.Patch(ctx, requestInstance, client.MergeFrom(originalReq))
+			if err != nil {
+				klog.Errorf("failed to remove finalizer for OperandRequest %s: %v", req.NamespacedName.String(), err)
+				return ctrl.Result{}, err
+			}
+		}
+		return ctrl.Result{}, nil
+	}
+
 	// Check if operator has the update permission to update OperandRequest
 	hasPermission := r.checkPermission(ctx, req)
 	if !hasPermission {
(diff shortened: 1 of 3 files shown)


horis233:
@Daniel-Fan @ZhuoxiLi Could you help review this PR?

controllers/operandrequest/operandrequest_controller.go
```diff
@@ -84,6 +84,28 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reconcileErr er
...
 	}()
 
+	// Remove finalizer when DeletionTimestamp none zero
+	if !requestInstance.ObjectMeta.DeletionTimestamp.IsZero() {
+
+		// Check and clean up the subscriptions
+		err := r.checkFinalizer(ctx, requestInstance)
+		if err != nil {
+			klog.Errorf("failed to clean up the subscriptions for OperandRequest %s: %v", req.NamespacedName.String(), err)
+			return ctrl.Result{}, err
```
Daniel-Fan, line 94: @horis233 Would you mind explaining it a little bit more? Why we move this section above the `add finalizer` would help with it?

From my understanding, if we directly delete the namespace, there are some other Common Service resources with the finalizer pending in the namespace, causing the situation that we could not get `err==nil` from `checkFinalizer` which means some resources cannot be deleted.

horis233: We move this logic before the permission check is because when the namespace is deleted, ODLM will lose the permission from the namespace scope operator. It can't pass the permission check. However, if there is nothing to delete or operand and operand are not in the deleting namespace, ODLM should be able to remove the finalizer from the operandrequest. That is the reason why we need this PR.

Daniel-Fan: Oh I see, usually the cloud pak create OperandRequest in the their own namespace, and the operators and operands are installed in `ibm-common-services` namespace, In this case, it could be correctly handled.

Thank you!

Daniel-Fan (approved):
/lgtm

controllers/operandrequest/operandrequest_controller.go
horis233: Yes, exactly

PR #693, opened 2021-06-15
fix requestNamespaces for iam/management-ingress bindinfo

ZhuoxiLi:

currently, delete zen instance, zen-tenant will not be removed from iam/management-ingress bindinfo's `requestNamespace`

### DIFF
--- a/controllers/operandbindinfo/operandbindinfo_controller.go
+++ b/controllers/operandbindinfo/operandbindinfo_controller.go
@@ -488,9 +488,7 @@ func (r *Reconciler) updateBindInfoPhase(bindInfoInstance *operatorv1alpha1.Oper
 	if bindInfoInstance.Status.Phase == phase && reflect.DeepEqual(requestNsList, bindInfoInstance.Status.RequestNamespaces) {
 		return
 	}
-	if len(requestNsList) != 0 {
-		bindInfoInstance.Status.RequestNamespaces = requestNsList
-	}
+	bindInfoInstance.Status.RequestNamespaces = requestNsList
 	bindInfoInstance.Status.Phase = phase
 }
 


Daniel-Fan (commented):
I am wondering whether we need to remove this?
https://github.com/IBM/operand-deployment-lifecycle-manager/blob/master/controllers/operandbindinfo/operandbindinfo_controller.go#L136
```
if len(requestNamespaces) == 0 {		
// There is no operand depend on the current bind info, nothing to do.		
      return ctrl.Result{}, nil	
}
```

I feel that if it is still there, we could not reach `r.updateBindInfoPhase()` in here
https://github.com/IBM/operand-deployment-lifecycle-manager/blob/master/controllers/operandbindinfo/operandbindinfo_controller.go#L206

ZhuoxiLi:
Hi @Daniel-Fan, I think we don't need to remove this if condition. I tested it and after removed zen instance, the `requestNamespaces` in bindinfo are removed successfully. 

 Also from code:
we assign `requestNamespaces`:
`requestNamespaces := registryInstance.Status.OperatorsStatus[bindInfoInstance.Spec.Operand].ReconcileRequests`

Here is an example of iam's `reconcileRequests`:
```
    ibm-iam-operator:
      reconcileRequests:
        - name: ibm-commonui-request
          namespace: ibm-common-services
        - name: common-service
          namespace: ibm-common-services
        - name: platform-api-request
          namespace: ibm-common-services
```

The above shows: in a normal cluster, it has some ns under `reconcileRequests` even without zen installed.  So I believe this part is not related to bindinfo.

Daniel-Fan:
Oh I see, thank you for the explanation!

Daniel-Fan (approved):
/lgtm

PR #748, opened 2021-09-27
add API for creating CR by OperandConfig

Daniel-Fan:
The previous design used `spec` section for the configuration of that specific resource.

But I am thinking that if we could make the resource creation more generic by using the `data` section instead of `spec`

Here is the example for `Job` resource:

Previously
```
kind: OperandConfig
metadata:
  name: operandconfig
  namespace: default
spec:
  services:
  - name: example-operator
     spec:
      example-operand:
        some-spec-field: myvalue
     resources:
     -   kind: Job
         name:  registry-key-generater
         apiVersion: batch/v1
         spec: ------------------- spec indicates the spec section in job, but other section could not be created
           template:
             metadata:
               ..         
```
Now
```
kind: OperandConfig
metadata:
  name: operandconfig
  namespace: default
spec:
  services:
  - name: example-operator
     spec:
      example-operand:
        some-spec-field: myvalue
     resources:
     -   kind: Job
         name:  registry-key-generater
         apiVersion: batch/v1
         data:    ------------------- data section indicates the configuration in the resource other than metadata, kind and apiVersion
           spec:  ------------------- spec indicates the spec section in job
             template:
               metadata:
               ..         
```

And for other resource like `Role`
```
kind: OperandConfig
metadata:
  name: operandconfig
  namespace: default
spec:
  services:
  - name: example-operator
     spec:
      example-operand:
        some-spec-field: myvalue
     resources:
     -   apiVersion: rbac.authorization.k8s.io/v1
         kind: Role
         namespace: default
         name: registry-key-generator
         data: 
           rules: ---------------------- Role resource does not have spec section, but it contains rules section
              - apiGroups: [""]
                resources: ["pods"]
                verbs: ["get", "watch", "list"]
            ......
```

### DIFF
--- a/api/v1alpha1/operandconfig_types.go
+++ b/api/v1alpha1/operandconfig_types.go
@@ -39,6 +39,34 @@ type ConfigService struct {
 	Spec map[string]runtime.RawExtension `json:"spec"`
 	// State is a flag to enable or disable service.
 	State string `json:"state,omitempty"`
+	// Resources is used to specify the k8s resources that are needed for the service.
+	// +optional
+	Resources []ConfigResource `json:"resources,omitempty"`
+}
+
+// ConfigResource defines the resource needed for the service
+type ConfigResource struct {
+	// Name is the resource name.
+	Name string `json:"name"`
+	// Kind identifies the kind of the custom resource.
+	Kind string `json:"kind"`
+	// APIVersion defines the versioned schema of this representation of an object.
+	APIVersion string `json:"apiVersion"`
+	// Namespace is the namespace of the resource.
+	// +optional
+	Namespace string `json:"namespace,omitempty"`
+	// Labels are the labels used in the resource.
+	// +optional
+	Labels map[string]string `json:"labels,omitempty"`
+	// Annotations are the annotations used in the resource.
+	// +optional
+	Annotations map[string]string `json:"annotations,omitempty"`
+	// Force is used to determine whether the existing resource should be overwritten.
+	// +optional
+	Force bool `json:"force,omitempty"`
+	// Data is the configuration map of custom resource.
+	// +optional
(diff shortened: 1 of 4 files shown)


api/v1alpha1/operandconfig_types.go
```diff
@@ -39,6 +39,34 @@ type ConfigService struct {
...
+	// Resources is used to specify the k8s resources that are needed for the service.
+	// +optional
+	Resources []ConfigResource `json:"resources,omitempty"`
+}
+
+// ConfigResource defines the resource needed for the service
+type ConfigResource struct {
+	// Name is the resource name.
+	Name string `json:"name"`
+	// Kind identifies the kind of the custom resource.
```
horis233, line 51: I believe `custom resource` is not accurate. How about `kubernetes resource`?

api/v1alpha1/operandconfig_types.go
```diff
@@ -39,6 +39,34 @@ type ConfigService struct {
...
+	// Labels are the labels used in the resource.
+	// +optional
+	Labels map[string]string `json:"labels,omitempty"`
+	// Annotations are the annotations used in the resource.
+	// +optional
+	Annotations map[string]string `json:"annotations,omitempty"`
+	// Force is used to determine whether the existing resource should be overwritten.
+	// +optional
+	Force bool `json:"force,omitempty"`
+	// Data is the configuration map of custom resource.
```
horis233, line 67: Same as above

api/v1alpha1/operandconfig_types.go
Daniel-Fan: Great Catch, updated!

horis233 (approved):
/lgtm

PR #749, opened 2021-09-28
create or update kubernetes resource by OperandConfig

Daniel-Fan:


controllers/operandrequest/reconcile_operand.go
```diff
@@ -179,6 +179,63 @@ func (r *Reconciler) reconcileOperand(ctx context.Context, requestInstance *oper
...
+				merr.Add(errors.Wrapf(err, "failed to get k8s resource %s/%s", k8sResNs, res.Name))
+			} else if apierrors.IsNotFound(err) {
+				if err := r.createK8sResource(ctx, k8sRes, res.Data.Raw, res.Labels, res.Annotations); err != nil {
+					merr.Add(err)
+				}
+			} else {
+				if checkLabel(k8sRes, map[string]string{constant.OpreqLabel: "true"}) {
+					// Update k8s resource
+					klog.V(3).Info("Found existing k8s resource: " + res.Name)
+					if err := r.updateK8sResource(ctx, k8sRes, res.Data.Raw, res.Labels, res.Annotations, res.Force); err != nil {
```
horis233, line 225: There is a point we need to discuss.

We expect `res.Force` by default to be true, but the default value of boolean is false, which will cause a misunderstanding if users don't set it or set it to false.

I am thinking if we can use the `force` in `*bool` type.

horis233: Also, could we add the check for `res.Force` at the same level as `checkLabel`? Then we don't need to include it in the update function

Daniel-Fan: Yeah, agree. I previously think the default value for `Force` is `false`. I will update the `*bool` type for `Force`.

> Could we add the check for res.Force at the same level as checkLabel? Then we don't need to include it in the update function

I am not sure it is reasonable...... If the `Force` is false, but the config of k8s resource is updated by users, in this case, we should still update the resources, right?

The path that we will not update the resource is either `checkLabel == false` or ` isEqual && Force == false`

Daniel-Fan: It seems that we could not use `*bool` as the type, because we [could not get the address of a constant.](https://stackoverflow.com/questions/42594789/initialize-string-pointer-in-struct)

Daniel-Fan:
@horis233 @ZhuoxiLi 
The PR is ready for merging. Please review it, thank you!

horis233 (approved):
/lgtm

PR #752, opened 2021-09-30
update the logic for updating the k8s resourcces

Daniel-Fan:
Some k8s resources could not be updated after it is created. `Job` for example

Since we will overwrite the resources when it is forced to update. now the update logic will delete the resource and then create a new one.

### DIFF
--- a/controllers/operandrequest/reconcile_operand.go
+++ b/controllers/operandrequest/reconcile_operand.go
@@ -32,6 +32,7 @@ import (
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	runtime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog"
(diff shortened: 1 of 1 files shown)


Daniel-Fan:
It is ready for merging.

controllers/operandrequest/reconcile_operand.go
```diff
@@ -805,89 +803,60 @@ func (r *Reconciler) createK8sResource(ctx context.Context, k8sResTemplate unstr
...
+	}, &existingK8sRes); err != nil {
+		return errors.Wrapf(err, "failed to get k8s resource -- Kind: %s, NamespacedName: %s/%s", kind, namespace, name)
+	}
 
+	if !checkLabel(existingK8sRes, map[string]string{constant.OpreqLabel: "true"}) {
+		return nil
+	}
+
+	// delete the existing k8s resource, some resources could not be updated after it has been created
+	if err := r.Client.Delete(ctx, &existingK8sRes); err != nil {
```
horis233, line 831: Will this line cause the k8s resource to get deleted in every reconcile loop?

Daniel-Fan: Yes, when ODLM is reconciling, and the resources is forced to update and label is correct. It will be deleted and re-created

ZhuoxiLi (approved):
/lgtm

controllers/operandrequest/reconcile_operand.go
Daniel-Fan: But even if we use the `Update()`, the resource will be updated in each reconcile loop except the resource could not be updated.

One solution is to compare the existing resource with the new template, and decide whether we should update it.

horis233: Could I ask which k8s resouce, that can't be updated?

Daniel-Fan: like `job`, some of fields in `template` is immutable, so we could not update it.

dram2040:
````
c840b9eee10088d767869a375b14c929d7743926
```
`

PR #755, opened 2021-10-12
Add the ability to delete the k8s resource

Daniel-Fan:


Daniel-Fan:
@horis233 
I found a logic defect when removing the k8s resources. I will update the code tomorrow.

Daniel-Fan:
Hi @horis233 @ZhuoxiLi 

Again the PR is ready for reviewing. Really sorry for taking such a long time~

Several stuff that needs to pat attention.

**The deletion logic is refactored and both `OperandRequest` and `OperandConfig` Controllers are involved.**

**Done by `OperandConfig` Controller** - When k8s resource templates are removed from `OperandConfig`, we will compare the current `OperandConfig` status with the template to delete the k8s resources.

**Done by `OperandRequest` Controller** - When its OperandRequest is removed, the k8s resource will be deleted based on the OperandConfig template.

~~Previously, the OperandConfig controller will not be triggered if the OperandRequest have been deleted. Now it is changed to always return true when there is a deletion of OperandRequest.~~
This part are not necessary to be changed anymore.

Add creation and deletion unit test for `OperandRequest` Controller.
And the complete e2e test for k8s resources is added as well.

Daniel-Fan:
@horis233 

Thank you for the reminder!
The unnecessary code has been removed, please review it again when you are available. Thank you!

PR #764, opened 2021-11-06
update golang to version 1.17.3 for EUS

Daniel-Fan:


go.mod
```diff
@@ -18,5 +18,60 @@ require (
 	sigs.k8s.io/controller-runtime v0.6.2
 )
 
+require (
```
horis233, line 21: Do we want to keep these libraries as old versions?

horis233 (approved):
/lgtm

PR #765, opened 2021-11-18
compare the hash of data section for Job resource to determine whether to update the Job template

Daniel-Fan:


horis233 (approved):
/lgtm

PR #766, opened 2021-11-23
Update the string splitter in OperandConfig status

Daniel-Fan:
Currently we use `.` to distinguish different segments in one string, for example
```
batch/v1.Job
```

But it does not work on some long API versions, for example
```
rbac.authorization.k8s.io/v1.Role
```

Update to `@` to avoid the repeating characters inside each segment.


horis233 (approved):
/lgtm

PR #774, opened 2022-01-10
check permission before creating k8s resource

Daniel-Fan:
When we try to create a configmap in `kube-public`, it is failed due to the lack of permission for ODLM
```
I0110 23:10:59.113598       1 reconcile_operand.go:241] ODLM doesn't have enough permission to reconcile k8s resource -- Kind: ConfigMap, NamespacedName: kube-public/game-demo
```

We will check operator permissions for both namespaced-scope and cluster-scope resources.

### DIFF
--- a/controllers/operandrequest/reconcile_operand.go
+++ b/controllers/operandrequest/reconcile_operand.go
@@ -30,11 +30,15 @@ import (
 
 	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/pkg/errors"
+	authorizationv1 "k8s.io/api/authorization/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	runtime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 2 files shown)


controllers/operandrequest/reconcile_operand.go
```diff
@@ -209,27 +213,32 @@ func (r *Reconciler) reconcileCRwithConfig(ctx context.Context, service *operato
 			k8sRes.SetName(res.Name)
 			k8sRes.SetNamespace(k8sResNs)
 
-			err := r.Client.Get(ctx, types.NamespacedName{
-				Name:      res.Name,
-				Namespace: k8sResNs,
-			}, &k8sRes)
+			verbs := []string{"create", "delete", "get", "update"}
```
horis233, line 216: @Daniel-Fan Do we need to check the permission for list and watch since we are using `client.get`?

Daniel-Fan: @horis233 Sorry for the late reply.....I did not notice the comment.

I created a ClusterRole and bind it to ODLM SA
```
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: odlm_configmap
rules:
  - verbs:
      - delete
      - update
      - create
      - get
    apiGroups:
      - ''
    resources:
      - configmaps
```
ODLM is able to create ConfigMap in other namespaces after the permission check.

Those verbs are enough to perform `Get`, `Update` and `Delete` from `Go-Client`.

horis233 (approved):
/lgtm

PR #804, opened 2022-03-18
Restart csv with exponential delay in checker

Daniel-Fan:
and remove Upgrade case in operator checker
and skip checking in Manual Approval Mode


horis233:
@Daniel-Fan Could you create another PR to disable this controller? Let's hold this PR until we verified all the possible cases.

Daniel-Fan:
@horis233 Of course, PR is open here https://github.com/IBM/operand-deployment-lifecycle-manager/pull/805

PR #796, opened 2022-03-09
fix: When operandrequest is empty then set OperandRequest to running …

horis233:
…as well

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

1. Please confirm that if this PR changes any image versions, then that's the sole change this PR makes.

**Release note**:

```release-note

```

PR #807, opened 2022-03-29
Revert "fix: When operandrequest is empty then set OperandRequest to running …"

horis233:
Reverts IBM/operand-deployment-lifecycle-manager#796


Daniel-Fan (approved):
/lgtm

PR #820, opened 2022-07-12
add version compare for subscription channel

Daniel-Fan:

When we deploy the same services from different OperandRegistries, the newer version(newer channel) of services will have higher priority and will be applied to the subscription.


qpdpQ (approved):

bitscuit (approved):
/lgtm

PR #877, opened 2023-01-06
Resolve version conflict in OperandRequest - Basic iteration

Daniel-Fan:

ODLM will properly handle the conflict to ensure NOT breaking existing services.

- When there are conflicting versions in multiple OperandRequests
  - keep the one that has been installed(install services at the version from first OperandRequest), do not upgrade or change it to avoid breaking anything.
     - **Future:** newer OperandRequests on different version of services created will be blocked.
     - **NOW:** newer OperandRequest will comply the existing service version and mark as `Running`.
- When existing version of services does not match with any versions in OperandRequests(OperandRequest/OperandRegistry has upgraded services to a newer version)
   - update service version to the minimal version requested from all OperandRequests


YCShen1010 (approved):
/lgtm

PR #888, opened 2023-01-31
add tombstone parameter in OperandRegistry

Daniel-Fan:
Introducing new parameter `.spec.operators[*].supportStatus` in OperandRegistry

`supportStatus` is the support status for services.
- When SupportStatus is `continous`, operator is supported to be fresh deployed via OperandRequest from scratch.
- When SupportStatus is `maintained`, operator is not supported to be fresh deployed via OperandRequest, only upgrade and deletion are allowed.
- The default value is `continous`

```
    - sourceNamespace: openshift-marketplace
      supportStatus: maintained  <-------------------
      channel: v3.23
      sourceName: opencloud-operators
      installPlanApproval: Automatic
      name: ibm-platform-api-operator
      packageName: ibm-platform-api-operator-app
      scope: public
      namespace: cs-2
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,12 @@ else
 ARTIFACTORYA_REGISTRY ?= "docker-na-public.artifactory.swg-devops.com/hyc-cloud-private-scratch-docker-local/ibmcom"
 endif
 
+ifdef DEV_REGISTRY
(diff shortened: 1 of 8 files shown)


YCShen1010 (approved):
/lgtm

PR #891, opened 2023-02-10
skip reconcile CR when it only exists in alm-example, not OperandConfig

Daniel-Fan:


YCShen1010 (approved):
/lgtm

PR #913, opened 2023-03-22
add preserve unknown fields to operandrequest status

bluzarraga:
add preserve unknown fields to status section for OperandRequest Status reporting


Daniel-Fan (approved):
/lgtm

PR #915, opened 2023-03-30
Remove annotation patch to avoid status lost

Daniel-Fan:
When controller patches the `metadata` section for OperandRequest, it actually overwrites its status section as well which is not expected.
Here is an example, when I add new `ibm-cert-manager-operator` into `.spec`, controller adds it into `requestInstance.status.members` section before patching.
However, this new added field in `status` disappears after patching `metadata` section.

```
I0330 02:40:45.011550 1 reconcile_operator.go:117] Before Patch, requestInstance status is [{ibm-im-operator {Running Running} []} {ibm-cert-manager-operator {Installing } []}]
I0330 02:40:45.011577 1 reconcile_operator.go:122] After Patch, requestInstance status is [{ibm-im-operator {Running Running} []}]
```

After some evaluations, I noticed that there is no really usage on this annotation `operator.ibm.com/operandregistry-is-not-found: 'false'`. If the `OperandRegistry` is not found, it be reflected from OperandRequest `status`.
As a result, I remove this annotation from OperandRequest in order to avoid any data loss on OperandRequest status which has highly impact on installation and uninstallation

### DIFF
--- a/api/v1alpha1/operandrequest_types.go
+++ b/api/v1alpha1/operandrequest_types.go
@@ -474,10 +474,10 @@ func (r *OperandRequest) setOperandReadyCondition(operandPhase ServicePhase, nam
 }
 
 // FreshMemberStatus cleanup Member status from the Member status list.
-func (r *OperandRequest) FreshMemberStatus(failedDeletedOperands *gset.Set) {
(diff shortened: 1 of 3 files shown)


bitscuit (approved):
/lgtm

PR #916, opened 2023-04-12
check for status.services, im service status

bluzarraga:
Issue 58155 has found that ODLM may not instantiate status.services on first pass through for reconciling an operandrequest. This pr adds a check to see if status.service exists and to see if im is ready. If either of these checks fails, the operandrequest is requeued.

I have been testing the changes with a 4.0 deployment and so far so good. I'm going to try cleaning it up before merging as it's a little ugly compared to the other checks here.

Testing:
- install bedrock 4.0, just cs operator and odlm
- replace odlm image with this one (can build with makefile or ask me and I'll send you an image)
- create operand request with just im
- watch im operand request, make sure it is eventually (in a relatively short time frame) updated to include status.services
- watch odlm logs to see if it's behaving as expected
- im opreq will create ibm-iam-request that has common ui and mongo, watch to make sure this opreq eventually has its status.service as well. It should not have the "waiting on im" log message in odlm

To rerun, you can delete the operandrequests, wait for everything to uninstall, then re-create the im operandrequest

Note: DO NOT restart the odlm operator after applying the operand request. It is a workaround in making sure the operand requests are reconciled to include status.service and could muddy the results of the test.

### DIFF
--- a/api/v1alpha1/operandrequest_types.go
+++ b/api/v1alpha1/operandrequest_types.go
@@ -495,6 +495,10 @@ func foundOperand(requests []Request, name string) bool {
 	return false
 }
 
+func (r *OperandRequest) operandRequested(name string) bool {
+	return foundOperand(r.Spec.Requests, name)
+}
+
 func getMemberStatus(status *OperandRequestStatus, name string) (int, *MemberStatus) {
 	for i, m := range status.Members {
 		if name == m.Name {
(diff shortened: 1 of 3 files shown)


Daniel-Fan:
The updates looks good to me. Let's go ahead merge it and unblock the issue.

On the other hand, I think we are trying to avoid hardcoded services in the code.
```
monitoredServices := []string{"ibm-iam-operator", "ibm-idp-config-ui-operator", "ibm-mongodb-operator", "ibm-im-operator"}
```

I have some new ideas, and I will open a new issue to have a team brainstorm on that.
- introduce some annotations in OperandConfig to mark monitored services - with backward compatibility
- introduce new spec parameter in OperandConfig to mark monitored services - without backward compatibility

PR #933, opened 2023-06-01
[LTSR] avoid OperandBindInfo label conflict

qpdpQ:
for ticket: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/58807
When two operandBindInfos are watching same secret/configmap, they are adding the same label to the secret/configmap ` operator.ibm.com/watched-by-opbi-with-name: ibm-licensing-bindinfo` 
This pr will try to fix it by adding a number after the label key

test:
1. install cs-operator and odlm
2. create two operandBindInfos for Licensing configmaps and secrets

expect result:
1. operandBindInfo will not fail
2. configmaps and secrets will be copied to another namespace

### DIFF
--- a/controllers/constant/constant.go
+++ b/controllers/constant/constant.go
@@ -37,6 +37,9 @@ const (
 	//OpbiNameLabel is the label used to add OperandBindInfo name to the secrets/configmaps watched by ODLM
 	OpbiNameLabel string = "operator.ibm.com/watched-by-opbi-with-name"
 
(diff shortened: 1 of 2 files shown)


qpdpQ:
this image can be used to verify: `quay.io/yuchen_li1/odlm-amd64@sha256:3101c350eb39aa9a1935797fe49297fff37bcfeea0263d3c9404ae0f7a9f09ba`

PR #927, opened 2023-05-10
Update test case

Daniel-Fan:
- Replace `etcd` operator by `jaeger` operator
- Replace `jenkins` operator by `mongodb-atlas` operator
- Update go module dependency to fit new operator packages
- Set OperandRequest as the OwnerReference of CR when it is directly created by OperandRequest(not via OperandConfig)
   - It is used to uninstall CRs when OperandRequest is removed.
- Ignore error on OperandRequest nor found when patching the OperandRequest status
   - When OperandRequest is deleted, no need to return patch error

### How to test
- Prepare a fresh cluster without anything installed
- Create following CatalogSource to include latest ODLM
```
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: odlm-catalog
  namespace: openshift-marketplace
spec:
  displayName: ODLM Test
  publisher: IBM
  sourceType: grpc
  image: quay.io/opencloudio/odlm-catalog:3357c1b2-dirty
  updateStrategy:
    registryPoll:
      interval: 45m
```
- Install ODLM from above new CatalogSource in `openshift-operators` namespace.
   - **NOTE: it is important to install ODLM in All Namespace Mode, it requires cluster permission to deploy operator in arbitrary namespace.**
```
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: ibm-odlm
  namespace: openshift-operators
spec:
  channel: v4.0
  installPlanApproval: Automatic
  name: ibm-odlm
  source: odlm-catalog
  sourceNamespace: openshift-marketplace
```
- From OCP console UI under `Installed Operators`, click into operator ODLM
- Select `OperandRegistry` tab, click `Create OperandRegistry`, click `Create` to create a default OperandRegistry template
```
kind: OperandRegistry
apiVersion: operator.ibm.com/v1alpha1
metadata:
  labels:
    app.kubernetes.io/instance: operand-deployment-lifecycle-manager
    app.kubernetes.io/managed-by: operand-deployment-lifecycle-manager
    app.kubernetes.io/name: operand-deployment-lifecycle-manager
  name: example-service
  namespace: openshift-operators
spec:
  operators:
    - channel: stable
      installMode: cluster
      name: jaeger
      namespace: default
      packageName: jaeger
      sourceName: community-operators
      sourceNamespace: openshift-marketplace
    - channel: stable
      name: mongodb-atlas-kubernetes
      namespace: default
      packageName: mongodb-atlas-kubernetes
[...]

PR #934, opened 2023-06-05
[LTSR] avoid OperandBindInfo label conflict

qpdpQ:
for ticket: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/58807
An easier way to all multiple operandBindInfos mounting same secret
When two operandBindInfos are watching same secret/configmap, they are adding the same label to the secret/configmap ` operator.ibm.com/watched-by-opbi-with-name: ibm-licensing-bindinfo` 
This pr will try to fix it by adding a number after the label key

test:
1. install cs-operator and odlm
2. create two operandBindInfos for Licensing configmaps and secrets

expect result:
1. operandBindInfo will not fail
2. configmaps and secrets will be copied to another namespace
3. there will be a new label in the secret `<namespace>.<name>/bindinfo: true`

This image can be used to verify `quay.io/yuchen_li1/odlm-amd64@sha256:ba111b9726675779bbdaafd6bfcbf583d6049d609686b70b3e4b43c0c285fcaa`

test looks good to me:
```
labels:
    ibm-common-services.ibm-iam-test-bindinfo/bindinfo: 'true'
    ibm-common-services.ibm-iam-bindinfo/bindinfo: 'true'
    operator.ibm.com/managedBy-opbi: original
```

### DIFF
--- a/controllers/operandbindinfo/operandbindinfo_controller.go
+++ b/controllers/operandbindinfo/operandbindinfo_controller.go
@@ -290,8 +290,7 @@ func (r *Reconciler) copySecret(ctx context.Context, sourceName, targetName, sou
 	}
 
 	ensureLabelsForSecret(secret, map[string]string{
-		constant.OpbiNsLabel:   bindInfoInstance.Namespace,
-		constant.OpbiNameLabel: bindInfoInstance.Name,
+		bindInfoInstance.Namespace + "." + bindInfoInstance.Name + "/bindinfo": "true",
 		constant.OpbiTypeLabel: "original",
 	})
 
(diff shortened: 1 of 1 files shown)


Daniel-Fan:
Thank you @qpdpQ !
It looks good to me!

I believe that this would be a good enhancement for Bedrock v4.1 as well.
Would you mind opening an enterprise Github issue to track this enhancement for Bedrock v4.1?

Also could you help to open an issue to track the E2E Test update on kind cluster for LTSR branch as well?
We could follow this PR to update the test case. https://github.com/IBM/operand-deployment-lifecycle-manager/pull/927
We could try `git cherry-pick` to pick up the test case changes. But some of the details would be different since there are some new enhancement in ODLM v4.0.

Daniel-Fan (approved):
/lgtm

PR #941, opened 2023-06-28
update grpc for cve

bluzarraga:


bluzarraga:
big jump in the grpc package so we need to make sure this build does not fail due to incompatibility between packages

Daniel-Fan (approved):
/lgtm

PR #952, opened 2023-07-24
update go-restful pkg for 3.19.14

bluzarraga:
go-restful package was included as a new dependency after last round of changes and it needs to be updated to address security vulnerabilities


YCShen1010 (approved):
/lgtm

PR #955, opened 2023-08-02
[4.2] Sort Catalogsource by priority in ODLM

YCShen1010:
for issue: https://github.ibm.com/ibmprivatecloud/roadmap/issues/54725
ODLM will prioritize CatalogSource for operator based on CatalogSource `.spec.priority` provided from OLM.

#### Current sequence of sorting order:

1. CatalogSources in the operator's namespace have a higher priority than those in other namespaces.
2. Within the same namespace, CatalogSources with higher priority values come first.
3. If both have the same namespace and priority, the CatalogSources are sorted based on their names in lexicographical order.

#### Test

1. test image: quay.io/yuchen_shen/odlm:catalog_prority
2. create `ibm-operator-catalog` in `openshift-marketplace` ns and give `.spec.priority` to `10`,  and same time create `cloud-native-postgresql-catalog` 
3. will see in sub `cloud-native-postgresql`, the catalogsource is `ibm-operator-catalog` as priority value 10 > 0(default)

### DIFF
--- a/controllers/operator/manager.go
+++ b/controllers/operator/manager.go
@@ -107,6 +107,7 @@ type CatalogSource struct {
 	Namespace         string
 	OpNamespace       string
 	RegistryNamespace string
+	Priority          int
 }
 
 type sortableCatalogSource []CatalogSource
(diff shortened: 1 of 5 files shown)


controllers/operator/manager.go
```diff
@@ -130,6 +124,12 @@ func (s sortableCatalogSource) Less(i, j int) bool {
 	if !inOpNsI && inOpNsJ {
 		return false
 	}
+	// Compare catalogsource priorities first, higher priority comes first
+	iPriority, jPriority := s[i].Priority, s[j].Priority
+	klog.Infof("iPriority: %v, jPriority: %v", iPriority, jPriority)
```
Daniel-Fan, line 129: Could we reveal this log information when verbosity is set to 2?

controllers/operator/manager.go
```diff
@@ -164,7 +164,11 @@ func (m *ODLMOperator) GetCatalogSourceFromPackage(ctx context.Context, packageN
 			if !channelCheck(channel, pm.Status.Channels) || (excludedCatalogSources != nil && util.Contains(excludedCatalogSources, pm.Status.CatalogSource)) {
 				continue
 			}
-			catalogSourceCandidate = append(catalogSourceCandidate, CatalogSource{Name: pm.Status.CatalogSource, Namespace: pm.Status.CatalogSourceNamespace, OpNamespace: namespace, RegistryNamespace: registryNs})
+			catalogsource := &olmv1alpha1.CatalogSource{}
+			if err := m.Reader.Get(ctx, types.NamespacedName{Name: pm.Status.CatalogSource, Namespace: pm.Status.CatalogSourceNamespace}, catalogsource); err != nil {
+				klog.Warning(err)
```
Daniel-Fan, line 169: I thought we could return error here.
Usually when `err != nil`, we will get `catalogsource` as `nil`. Therefore, it will error out in the next line when we try to read `catalogsource.Spec.Priority`

Daniel-Fan (changes requested):

Daniel-Fan (approved):
/lgtm

PR #1002, opened 2023-12-05
Enhance error handling for uninstall procedure

Daniel-Fan:
To address issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/61390

Make several enhancements:
1. Increase the number of OperandRequests which could be reconciled by ODLM simultaneously to 10
2. Better Error handling to avoid removing operand from remainingOperands set, to keep track of failed uninstallation
3. In parent OperandRequest uninstallation reconciliation, skip waiting for its child OperandRequest to be deleted when this child OperandRequest is a "internal" OperandRequest as the operand of services.
   - For example, `keycloak-operator` has a "internal" OperandRequest requesting `edb-keycloak` as the dependency of Keycloak operator.


YCShen1010 (approved):
/lgtm

PR #1010, opened 2024-01-11
Allow resources to be updated by ODLM when control label is specified

Daniel-Fan:
When the label `operator.ibm.com/opreq-control: 'true'` for a resource is specified in OperandConfig, ODLM will always update this resources.

This is used to handle an upgrade scenario where the resource is previously not managed by ODLM, but it is managed by ODLM after upgrade.


bitscuit (approved):
/lgtm

PR #1016, opened 2024-02-20
Prioritize the Catalog which ODLM is deployed from

Daniel-Fan:
issue: https://github.ibm.com/ibmprivatecloud/roadmap/issues/62275

Add an additional layer on sorting the Catalog
1. CatalogSources in operator namespace (private CatalogSource) have a higher priority than those in other namespaces (global CatalogSource).
2. CatalogSources with higher `.spec.priority` values will be picked up first.
3. **CatalogSource that the cs-operator/odlm is already installed from would be picked up next, if it is in the list of CatalogSources for the operator we want to install**
    - If this CatalogSource contains IM operator. i.e. `opencloud-operators`, it would be picked first instead of the `ibm-im-catalog` in lexicographical order.
    - If this CatalogSource does not contain IM operator. i.e., `ibm-installer-catalog`, it would not even in the list of Catalog used for IM installation.
4. If both have the same namespace and priority, the CatalogSources are sorted based on their names in lexicographical order. 

### Test with `opencloud-operator` and `ibm-operator-catalog` Catalog

- Prepare two CatalogSource `opencloud-operator` and `ibm-operator-catalog`
- Install CS from `opencloud-operator` Catalog
- Update ODLM CSV with image `quay.io/opencloudio/odlm:6994c280-dirty`
- Create a OperandRequest, requesting `ibm-im-operator`
- Check `ibm-im-operator`, the subscription `.spec.catalogName: opencloud-operators`

Although `ibm-operator-catalog` has higher priority than `opencloud-operator` in lexicographical order, ODLM is installed from `opencloud-oeprator` catalog. As a result, ODLM installs IAM from `opencloud-operator` catalog over `ibm-operator-catalog`

```
I0220 19:55:24.915973 1 manager.go:208] Found 2 CatalogSources for PackageManifest ibm-iam-operator in the namespace 4-5-test has channel v4.4
I0220 19:55:24.915998 1 manager.go:212] The 0th sorted CatalogSource is opencloud-operators in namespace openshift-marketplace with priority: 0
I0220 19:55:24.916002 1 manager.go:212] The 1th sorted CatalogSource is ibm-operator-catalog in namespace openshift-marketplace with priority: 0
```

### DIFF
--- a/controllers/operator/manager.go
+++ b/controllers/operator/manager.go
@@ -76,6 +76,20 @@ func (m *ODLMOperator) GetOperandRegistry(ctx context.Context, key types.Namespa
 	if reg.Annotations != nil && reg.Annotations["excluded-catalogsource"] != "" {
 		excludedCatalogSources = strings.Split(reg.Annotations["excluded-catalogsource"], ",")
 	}
+	// Get catalog used by ODLM itself by check its own subscription
+	opts := []client.ListOption{
+		client.MatchingLabels{fmt.Sprintf("operators.coreos.com/ibm-odlm.%s", util.GetOperatorNamespace()): ""},
+		client.InNamespace(util.GetOperatorNamespace()),
+	}
+	odlmCatalog := ""
+	odlmCatalogNs := ""
+	odlmSubList := &olmv1alpha1.SubscriptionList{}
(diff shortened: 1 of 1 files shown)


YCShen1010 (approved):
/lgtm

PR #1019, opened 2024-02-27
Enable deep JSON merging for Kubernetes resources when updating them

YCShen1010:
Issue: https://github.ibm.com/ibmprivatecloud/roadmap/issues/61283

### Context 
For this enhancement, we allow users to perform additional customizations on Kubernetes resources created ODLM. ODLM will compare the existing Custom Resource in the cluster with the template in the OperandConfig, field by field. It will retain the fields that are not defined in OperandConfig instead of overwriting them.

### How to test
Test image: quay.io/yuchen_shen/odlm:deep_merge

#### CASE I: Test for map-type values
**resource: `keycloak-edb-cluster` Cluster CR**
Before apply the image 
- Remove `description` section under `.data.spec` for `keycloak-edb-cluster` service in OperandConfig, and restart ODLM pod, will observe the `description: EDB Database with IBM License Key` will be removed from Cluster CR

After apply the test image
- Add `description: EDB Database with IBM License Key` back to existing Cluster CR, keep description section removed in the OperandConfig, and restart ODLM pod, will observe the description will be retained in `keycloak-edb-cluster` CR

#### CASE II: Test for slice-type values
**resource: `edb-license-role` Role CR**
After apply the image 
- Adding or deleting rules to Role CR and restarting the ODLM pod will result in the observation that the newly added or deleted rules revert back to exactly what is specified in OperandConfig.
- Adding a new set of rules, which is not defined in OperandConfig, it will keep the new rules in the role.

### DIFF
--- a/controllers/operandrequest/reconcile_operand.go
+++ b/controllers/operandrequest/reconcile_operand.go
@@ -28,6 +28,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/mohae/deepcopy"
 	olmv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/pkg/errors"
 	authorizationv1 "k8s.io/api/authorization/v1"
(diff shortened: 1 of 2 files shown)


controllers/operandrequest/reconcile_operand.go
```diff
@@ -1057,61 +1057,66 @@ func (r *Reconciler) updateK8sResource(ctx context.Context, existingK8sRes unstr
...
 		}
 
-		// isEqual := r.CheckAnnotation(existingK8sRes, newAnnotations) && r.CheckLabel(existingK8sRes, newLabels)
 		if k8sResConfig != nil {
+
 			k8sResConfigDecoded := make(map[string]interface{})
 			k8sResConfigUnmarshalErr := json.Unmarshal(k8sResConfig.Raw, &k8sResConfigDecoded)
 			if k8sResConfigUnmarshalErr != nil {
 				klog.Errorf("failed to unmarshal k8s Resource Config: %v", k8sResConfigUnmarshalErr)
 			}
```
Daniel-Fan, line 1067: I believe we could remove this section.

controllers/operandrequest/reconcile_operand.go
```diff
@@ -1057,61 +1057,66 @@ func (r *Reconciler) updateK8sResource(ctx context.Context, existingK8sRes unstr
...
-		if err := r.setOwnerReferences(ctx, &existingK8sRes, ownerReferences); err != nil {
-			return false, errors.Wrapf(err, "failed to set ownerReferences for k8s resource -- Kind: %s, NamespacedName: %s/%s", kind, namespace, name)
-		}
+			if reflect.DeepEqual(existingK8sRes.Object, updatedExistingK8sRes) {
+				return true, nil
+			}
+			klog.Infof("updating k8s resource with apiversion: %s, kind: %s, %s/%s", apiversion, kind, namespace, name)
 
-		klog.V(2).Infof("updating k8s resource with apiversion: %s, kind: %s, %s/%s", apiversion, kind, namespace, name)
+			existingK8sRes.Object = updatedExistingK8sRes
```
Daniel-Fan, line 1089: It seems that we are not `ensureAnnotation`, `ensureLabel` and `setOwnerReferences` on `updatedObject`. Currently it is applying to `existingK8sRes`

We might need to change the sequence here
1. deep copy a original version of `existingK8sRes.Object` - does not exists now.
2. `existingK8sRes.Object = updatedExistingK8sRes`
3. Ensure annotation, label and OwnerReference on new `updatedObject`
4. Check if new assigned object is deep equal with original one.

Daniel-Fan (changes requested):

Daniel-Fan (approved):
/lgtm

PR #1026, opened 2024-03-19
updated base image to explicitly use docker.io instead of implied

bitscuit:
otherwise error occurs when using podman and docker.io has not been configured as default registry


Daniel-Fan (approved):
/lgtm

PR #1028, opened 2024-03-25
Support uninstallation on intra-dependency

Daniel-Fan:
uninstallation issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/62243

### Context
There are several services depending on the same operator, but different operands.

For example, both IM service and Keycloak service depend on the same EDB operator. However, each of them has different EDB Cluster CR. As a result, we are using different entires in OperandRegistry/OperandConfig to distinguish EDB used by IM as `common-service-postgresql`, and EDB used by Keycloak as `edb-keycloak`

For simplicity, I will only use EDB Cluster CR as the operand created by ODLM.

- If we request `common-service-postgresql`, it will install EDB operator, and EDB Cluster CR `common-service-db`.
- If we request `edb-keycloak`, it will install EDB operator, and EDB Cluster CR `keycloak-edb-cluster`.

When we request both of them in the OperandRequest, they share the same EDB operator, but different EDB Cluster CRs.

During the uninstallation, following pattern should be fulfilled.
- EDB operator would be uninstalled only when there is no `common-service-postgresql` and `edb-keycloak` requested
- `common-service-db` Cluster CR should be removed as long as there is no `common-service-postgresql` in OperandRequest
- `keycloak-edb-cluster` Cluster CR should be removed as long as there is no `edb-keycloak` in OperandRequest

### How to test
- Install CS 4.6 daily build in one namespace
- Update ODLM CSV to replace image to `quay.io/opencloudio/odlm:9c23dfda`, and update ImagePullPolicy to `Always`
- Create an OperandRequest to request both `common-service-postgresql` and `edb-keycloak`
    ```yaml
    kind: OperandRequest
    metadata:
      name: example-service
    spec:
      requests:
        - operands:
            - name: common-service-postgresql
            - name: edb-keycloak
          registry: common-service
    ```
- Observe that there are `common-service-db-1` and `keycloak-edb-cluster-1` pods available, representing two EDB Cluster CR respectively.
- Observe that there is a `postgresql-operator-controller-manager` pod available, representing EDB operator
- Update OperandRequest, to remove `common-service-postgresql`
- Observe that there is NO `common-service-db-1` pod in the namespace.
- Observe that there is still `keycloak-edb-cluster-1` and `postgresql-operator-controller-manager` pod in the namespace.
- Delete the entire OperandRequest
- Observe that there is no more EDB related pod left in the namespace.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,9 +100,6 @@ BUNDLE_DEFAULT_CHANNEL := --default-channel=$(DEFAULT_CHANNEL)
 endif
 BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 
-# Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true"
-
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
 GOBIN=$(shell go env GOPATH)/bin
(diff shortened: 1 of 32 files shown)


YCShen1010 (approved):
/lgtm

PR #1039, opened 2024-04-22
Update ODLM module to match major version v4

Daniel-Fan:
https://go.dev/ref/mod#major-version-suffixes
Starting with major version 2, module paths must have a major version suffix like /v2 that matches the major version. For example, if a module has the path example.com/mod at v1.0.0, it must have the path example.com/mod/v2 at version v2.0.0.


YCShen1010 (approved):
/lgtm

PR #1048, opened 2024-05-27
Only discover catalog for an operator when it has been requested in OperandRequest

Daniel-Fan:

After comparing the logs in ODLM from CPFS 4.5 and 4.7 release, the main difference happened on `reconciling Operators for OperandRequest`. And the root cause is about [removing hardcode Catalog specified in the OperandRegistry](https://github.com/IBM/ibm-common-service-operator/pull/1802) for the adoption of Catalog separation.

### Context
After hardcode Catalog specified in OperandRegistry is removed, ODLM needs to do discovery for each operator in OperandRegistry, and find their available Catalog for installation. There are almost 50 operators in OperandRegistry, and it takes about 1 mins to find the Catalog for all of them.

### Solution
However, from the operational perspective, ODLM does not have to find the CatalogSource for each operator in OperandRegistry since not all of them are requested in OperandRequest.
We will need to refactor the code, to only find the Catalog for the operators that are requested in OperandRequest. So it will significantly reduce the time on `reconciling Operators for OperandRequest`

Fixes https://github.ibm.com/IBMPrivateCloud/roadmap/issues/63499#issuecomment-80082231

### DIFF
--- a/controllers/operandbindinfo/operandbindinfo_controller.go
+++ b/controllers/operandbindinfo/operandbindinfo_controller.go
@@ -159,11 +159,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 		return ctrl.Result{}, nil
 	}
 	// Get the operand namespace
-	operandOperator := registryInstance.GetOperator(bindInfoInstance.Spec.Operand)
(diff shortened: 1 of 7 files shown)


YCShen1010 (approved):
/lgtm

PR #1050, opened 2024-06-03
Skip catalog discover when operator is not found in OperandRegistry

Daniel-Fan:
When operator is not found in OperandRegistry, `opt` is `nil` and there should be no catalog discovery for this `nil` object.

Fixes https://github.ibm.com/IBMPrivateCloud/roadmap/issues/63690#issuecomment-81218953

### Test
See comment https://github.ibm.com/IBMPrivateCloud/roadmap/issues/63690#issuecomment-81222501

### DIFF
--- a/controllers/operator/manager.go
+++ b/controllers/operator/manager.go
@@ -402,6 +402,10 @@ func (m *ODLMOperator) GetOperatorNamespace(installMode, namespace string) strin
 // GetOperandFromRegistry gets the Operand from the OperandRegistry
 func (m *ODLMOperator) GetOperandFromRegistry(ctx context.Context, reg *apiv1alpha1.OperandRegistry, operandName string) (*apiv1alpha1.Operator, error) {
 	opt := reg.GetOperator(operandName)
+	if opt == nil {
+		return nil, nil
+	}
+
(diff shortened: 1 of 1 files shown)


controllers/operator/manager.go
```diff
@@ -402,38 +402,40 @@ func (m *ODLMOperator) GetOperatorNamespace(installMode, namespace string) strin
...
-		odlmCatalog = odlmSubList.Items[0].Spec.CatalogSource
-		odlmCatalogNs = odlmSubList.Items[0].Spec.CatalogSourceNamespace
-	}
-
-	if opt.SourceName == "" || opt.SourceNamespace == "" {
-		catalogSourceName, catalogSourceNs, err := m.GetCatalogSourceFromPackage(ctx, opt.PackageName, opt.Namespace, opt.Channel, reg.Namespace, odlmCatalog, odlmCatalogNs, excludedCatalogSources)
-		if err != nil {
-			return nil, err
-		}
+	if opt != nil {
```
bitscuit, line 405: suggest to invert the condition, so that if opt is nil, then return nil. Will be easier to read the code, and best practice is to return as early as possible.

bitscuit (approved):
/lgtm

PR #1053, opened 2024-06-05
reconcile k8s resource in parallel and with retry logic

Daniel-Fan:
Reduce the k8s resource reconciliation time

Fixes https://github.ibm.com/IBMPrivateCloud/roadmap/issues/63499

### Description
ODLM will be capable of reconciling 3 resources(resource A, B, C) in parallel, to speed up the overall reconciliation time.

#### Reconcile the same resource with retry
It is possible that the resource C depends on certain fields in resource A, so the reconciliation of resource C can't be completed until resource A is completed.
In order to minimize the waiting time, ODLM will always re-try the resource A reconciliation several times before raising an error. And the waiting time for each re-try is increased exponentially to avoid heavy workload on APIServer.
```golang
for i := 0; i < chunkSize; i++ {
    err = r.reconcileK8sResource(ctx, res, serviceName, opConfigName, opConfigNs)
    if err == nil {
	    return nil
    }
    klog.Errorf("Failed to reconcile k8s resource %s/%s: %v", res.Namespace, res.Name, err)
    if i < chunkSize-1 {
	    waitTime := time.Duration((1 << i) * 4 * int(time.Second))
	    klog.Warningf("Retry reconcile k8s resource %s/%s after %v", res.Namespace, res.Name, waitTime)
	    time.Sleep(waitTime)
    }
} 
```

#### Remove resource deep equal comparison
Previously we always have a code `DeepEqual(oldObject, newObject)` to check if the object is updated or not before we put a `Update` command to APIServer.
However, I noticed that the `DeepEqual(oldObject, newObject)` always returns `false` in the practice even there is no object update - not sure why it happens.

As a result, I remove this `DeepEqual` comparison to reduce runtime, and I always put a `Update` command to APIServer and rely on k8s `resourceVersion` to check if object is truly updated or not.
```golang
resourceVersion := existingK8sRes.GetResourceVersion()
err = r.Update(ctx, &existingK8sRes)
...
if UpdatedK8sRes.GetResourceVersion() != resourceVersion {
   klog.Infof("Finish updating the k8s Resource: -- Kind: %s, NamespacedName: %s/%s", kind, namespace, name)
} else {
   klog.Infof("No updates on k8s resource with apiversion: %s, kind: %s, %s/%s", apiversion, kind, namespace, name)
}
```

#### Other trivial updates
- Update the trigger condition on reconciliation for OperatorConfig

### DIFF
--- a/controllers/constant/constant.go
+++ b/controllers/constant/constant.go
@@ -90,4 +90,7 @@ const (
 
 	//DefaultCSVWaitPeriod is the default period for wait CSV ready
 	DefaultCSVWaitPeriod = 1 * time.Minute
+
+	//DefaultCRRetryNumber is the default maximum number of retry for reconciling a custom resource
+	DefaultCRRetryNumber = 3
 )
--- a/controllers/operandrequest/reconcile_operand.go
+++ b/controllers/operandrequest/reconcile_operand.go
@@ -236,7 +236,7 @@ func (r *Reconciler) reconcileCRwithConfig(ctx context.Context, service *operato
 			go func(res operatorv1alpha1.ConfigResource) {
 				defer wg.Done()
 				defer func() { <-semaphore }() // release semaphore
-				err := r.reconcileK8sResourceWithRetries(ctx, res, service.Name, opConfigName, opConfigNs, chunkSize)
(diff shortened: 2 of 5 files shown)


bitscuit (approved):
/lgtm

PR #1078, opened 2024-08-28
added feature to handle user managed operators (BYO)

bitscuit:


controllers/operandrequest/reconcile_operator.go
```diff
@@ -160,6 +160,22 @@ func (r *Reconciler) reconcileSubscription(ctx context.Context, requestInstance
...
+		sub.Annotations[requestInstance.Namespace+"."+requestInstance.Name+"."+operand.Name+"/operatorNamespace"] = namespace
+		if err = r.updateSubscription(ctx, requestInstance, sub); err != nil {
+			requestInstance.SetMemberStatus(opt.Name, operatorv1alpha1.OperatorFailed, "", mu)
+			return err
+		}
+		requestInstance.SetMemberStatus(opt.Name, operatorv1alpha1.OperatorUpdating, "", mu)
+		// requestInstance.SetMemberStatus(operand.Name, operatorv1alpha1.OperatorRunning, operatorv1alpha1.ServiceRunning, mu)
+		return nil
+	}
+
```
Daniel-Fan, line 178: Hi @bitscuit 

I am wondering if we are able to handle the case where user creates subscription in `openshift-operators` namespace with all namespaces mode, and install CPFS in own namespace mode.

In this case, we will not seeing any subscription in CPFS tenant.

Daniel-Fan: I am thinking it is difficult to handle the uninstallation at this moment since most of our uninstallation is based on subscription annotations.

I feel that if we could not solve it in 4.6.6 release, we might consider it as known issue on uninstallation when `user managed` feature is on.

Daniel-Fan: I am facing the same uninstallation with fallback channels, I am seeing if I could refactor the logic in uninstallation section.

controllers/operandrequest/reconcile_operand.go
```diff
@@ -154,26 +168,28 @@ func (r *Reconciler) reconcileOperand(ctx context.Context, requestInstance *oper
...
-			var requestList []string
-			reg, _ := regexp.Compile(`^(.*)\.(.*)\.(.*)\/request`)
-			for anno, version := range sub.Annotations {
-				if reg.MatchString(anno) && version == sub.Spec.Channel {
-					requestList = append(requestList, anno)
+			// if err := r.DeleteRedundantCSV(ctx, csv.Name, csv.Namespace, registryKey.Namespace, opdRegistry.PackageName); err != nil {
+			// 	merr.Add(errors.Wrapf(err, "failed to delete the redundant ClusterServiceVersion %s in the namespace %s", csv.Name, csv.Namespace))
+			// 	requestInstance.SetMemberStatus(operand.Name, operatorv1alpha1.OperatorFailed, "", &r.Mutex)
+			// 	continue
+			// }
```
bitscuit, line 175: I'm not really sure why this DeleteRundantCSV is needed when reconciling operands. Commented it out because it interferes with trying to reconcile OperandRequests purely using CSVs (no Subscription info)

controllers/operandrequest/reconcile_operator.go
bitscuit: done

Daniel-Fan (approved):
/lgtm

PR #1083, opened 2024-09-25
Revert the host back to default value

YCShen1010:
**Which issue(s) this PR fixes** 
Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/64018

This PR addresses the need to handle custom `spec.host` values for OpenShift Route resources when creating it. Specifically, if the resource is Route, the `spec.host value` is now dynamically read from the provided OperandConfig template. A hash of the `spec.host` value is then calculated and added to the resource's annotations. If the custom host is empty from Opcon, then the host value from the Route will be reverted back to default

### DIFF
--- a/controllers/operandrequest/reconcile_operand.go
+++ b/controllers/operandrequest/reconcile_operand.go
@@ -1010,9 +1010,21 @@ func (r *Reconciler) createK8sResource(ctx context.Context, k8sResTemplate unstr
 		if k8sResConfigUnmarshalErr != nil {
 			klog.Errorf("failed to unmarshal k8s Resource Config: %v", k8sResConfigUnmarshalErr)
 		}
-
 		for k, v := range k8sResConfigDecoded {
 			k8sResTemplate.Object[k] = v
+		}
+
+		if kind == "Route" {
+			if host, found := k8sResConfigDecoded["spec"].(map[string]interface{})["host"].(string); found {
+				hostHash := util.CalculateHash(host)
+
+				if newAnnotations == nil {
+					newAnnotations = make(map[string]string)
+				}
+				newAnnotations["operator.ibm.com/odlm.route.hashedData"] = hostHash
+			} else {
+				klog.Warningf("spec.host not found in Route %s/%s", namespace, name)
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


controllers/operandrequest/reconcile_operand.go
```diff
@@ -948,16 +948,48 @@ func (r *Reconciler) createK8sResource(ctx context.Context, k8sResTemplate unstr
...
+
+		// Calculate and set hash for spec.host
+		hostHash := util.CalculateHash(host)
+		if newAnnotations == nil {
+			newAnnotations = make(map[string]string)
+		}
+		newAnnotations["openshift.io/keycloak-host.hashedData"] = hostHash
+		// klog.Infof("22222 Set host hash annotation: %s", hostHash)
+	}
+
```
Daniel-Fan, line 968: I am wondering that if this is redundant?

It seems that variable `k8sResTemplate` only contains kind, name and namespace, and it does not have any `.spec` field yet.

controllers/operandrequest/reconcile_operand.go
```diff
@@ -948,16 +948,48 @@ func (r *Reconciler) createK8sResource(ctx context.Context, k8sResTemplate unstr
...
 		}
+
+		if kind == "Route" {
+			if host, found := k8sResConfigDecoded["spec"].(map[string]interface{})["host"].(string); found {
+				hostHash := util.CalculateHash(host)
+
+				if newAnnotations == nil {
+					newAnnotations = make(map[string]string)
+				}
+				newAnnotations["openshift.io/keycloak-host.hashedData"] = hostHash
```
Daniel-Fan, line 987: can we use odlm specific label? For example `operator.ibm.com/odlm.route.hashedData`

controllers/operandrequest/reconcile_operand.go
```diff
@@ -948,16 +948,48 @@ func (r *Reconciler) createK8sResource(ctx context.Context, k8sResTemplate unstr
...
+				klog.Infof(" 2222  Route spec.host: %s, calculated host-hash: %s", host, hostHash)
+			} else {
+				klog.Warningf("spec.host not found in k8sResConfig for Route")
```
Daniel-Fan, line 990: Could we improve the warning message to point out the Route namespace and name?

controllers/operandrequest/reconcile_operand.go
```diff
@@ -1038,6 +1071,63 @@ func (r *Reconciler) updateK8sResource(ctx context.Context, existingK8sRes unstr
 		return nil
 	}
 
+	if kind == "Route" {
```
Daniel-Fan, line 1074: I am wondering if we are able to wrap the `update` action for `Job` and `Route` into their own functions?
It would help to reduce the number of lines for the function `updateK8sResource`

Daniel-Fan (approved):
/lgtm

PR #1084, opened 2024-09-26
Use PackageName name as the name of subscription

YCShen1010:
This fix will ensure that no duplicate subscriptions are created based on the same operator package name and different channels. Instead, if there are two OperandRequests with the same operator but different channels, ODLM will create only one subscription with the package name, as only one operator/subscription with a single channel is allowed to be installed in one tenant.

**Which issue(s) this PR fixes** 
Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/64818

**Before/After fix in**
Request two same operator with diff channel
```
kind: OperandRequest
metadata:
  name: example-service
spec:
  requests:
    - operands:
        - name: cloud-native-postgresql-v1.22
        - name: cloud-native-postgresql
      registry: common-service
```

Before fix will see fail to create two subs:
Screenshot 2024-09-25 at 21 50 59

After fix, successfully created one sub:
Screenshot 2024-09-25 at 22 25 16

### DIFF
--- a/controllers/operandrequest/operandrequest_controller.go
+++ b/controllers/operandrequest/operandrequest_controller.go
@@ -100,6 +100,9 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 		if reflect.DeepEqual(existingInstance.Status, requestInstance.Status) {
 			return
 		}
+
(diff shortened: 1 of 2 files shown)


Daniel-Fan (approved):
/lgtm

Daniel-Fan (approved):
/lgtm

PR #1086, opened 2024-09-28
Enhance ODLM create/update logic by adding hash values

YCShen1010:
Related issues: 
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/64018
https://github.ibm.com/IBMPrivateCloud/roadmap/issues/64893

### Context

- Every object created by ODLM adds a hash annotation, which is according to the template defined in OperandConfig.
- Introduced a hash-based comparison to determine whether updates are needed.
- For non-Job resources, the template is merged with the existing resource if the hashes match, ensuring new fields are added without overwriting unchanged ones. If the hashes don't match, the resource is overwritten by the template.
- Implemented a solution for deep merging resources, including metadata and spec, by using the merge functions.

### DIFF
--- a/controllers/constant/constant.go
+++ b/controllers/constant/constant.go
@@ -64,6 +64,12 @@ const (
 	//HashedData is the key for checking the checksum of data section
 	HashedData string = "hashedData"
 
+	//HashedData is the key for k8s Resource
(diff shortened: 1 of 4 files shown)


Daniel-Fan (approved):
/lgtm

PR #1092, opened 2024-10-10
label length limitation

qpdpQ:
What this PR does / why we need it:
k8s label name length limit to 64 characters, so olm will cut off any label longer than 64 characters
We rely on label to find csv and subscription, so we also need to cut of the label if it is longer than 64 characters.
Which issue(s) this PR fixes:
script enhancement for # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/64378


Daniel-Fan (approved):
/lgtm

PR #1097, opened 2024-11-15
Add a cluster permission check before deleting v3 resources

YCShen1010:
**What this PR does / why we need it**: Added logic to check if the permissions of v3 resources `certmanagers` and `auditloggings` exist and only execute get and delete operations if they are present

**Which issue(s) Fixes** https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65322

### How to test
1. Install a v3 operator and request `ibm-auditlogging-operator` and `ibm-cert-manager` (not using the latest LTSR version)
2. Isolate upgrade to v4, and remove these two cluster permission for ODLM in clusteRole and CSV
3. Delete the OperandRequest, will see the permission failed error
4. Apply the test image `quay.io/yuchen_shen/odlm:remove_cluster_role` and the error would be gone, and CRs are not deleted.


Daniel-Fan (approved):
/lgtm

PR #1099, opened 2024-11-26
Discover private CatalogSource only if no catalog cluster permission exists

YCShen1010:
Issue: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65336

### Context
Added permission check function to verify if it has the required permission, and then read catalog source by checking it before attempting to fetch them. If lacks permissions, limit catalog source processing to those in the operator namespace only.

### How to test
Test image: quay.io/yuchen_shen/odlm:catalog_permission

**Case I:** Apply private CatalogSource for ODLM, IM and EDB
1. Install v4 operator and remove the catalog cluster permission for ODLM
2. IM and EDB could be installed successfully

**Case II:** Apply CatalogSource as usual in `openshift-marketplace`
1.  Install v4 operator and remove the catalog cluster permission for ODLM
2. Request to install IM, will get error msg 
    ```failed to reconcile Operators for OperandRequest non-pravite-catalog/example-service: the following errors occurred:
    - failed to find catalogsource for subscription openshift-marketplace/ibm-iam-operator
    ```
**Case III**: Apply CatalogSource as usual in `openshift-marketplace`
1.  Install v4 operator and keep the cluster permission
2. Request to install IM, everything will be installed

### DIFF
--- a/controllers/operator/manager.go
+++ b/controllers/operator/manager.go
@@ -27,6 +27,7 @@ import (
 	operatorsv1 "github.com/operator-framework/operator-lifecycle-manager/pkg/package-server/apis/operators/v1"
 	"github.com/pkg/errors"
 	"golang.org/x/mod/semver"
+	authorizationv1 "k8s.io/api/authorization/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -172,11 +173,19 @@ func (m *ODLMOperator) GetCatalogSourceAndChannelFromPackage(ctx context.Context
 			if excludedCatalogSources != nil && util.Contains(excludedCatalogSources, pm.Status.CatalogSource) {
 				continue
 			}
+
+			hasCatalogPermission := m.CheckResAuth(ctx, namespace, "operators.coreos.com", "catalogsources", "get")
+			if !hasCatalogPermission {
+				klog.V(2).Infof("No permission to get CatalogSource %s in the namespace %s", pm.Status.CatalogSource, pm.Status.CatalogSourceNamespace)
+				continue
+			}
+			// Fetch the CatalogSource if cluster permission allows
 			catalogsource := &olmv1alpha1.CatalogSource{}
 			if err := m.Reader.Get(ctx, types.NamespacedName{Name: pm.Status.CatalogSource, Namespace: pm.Status.CatalogSourceNamespace}, catalogsource); err != nil {
 				klog.Warning(err)
 				continue
 			}
+
 			currentCatalog := CatalogSource{
 				Name:                 pm.Status.CatalogSource,
 				Namespace:            pm.Status.CatalogSourceNamespace,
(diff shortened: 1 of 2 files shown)


controllers/operator/manager.go
```diff
@@ -172,11 +173,22 @@ func (m *ODLMOperator) GetCatalogSourceAndChannelFromPackage(ctx context.Context
...
+				// Fetch the CatalogSource if cluster permission allows
+				catalogsource := &olmv1alpha1.CatalogSource{}
+				if err := m.Reader.Get(ctx, types.NamespacedName{Name: pm.Status.CatalogSource, Namespace: pm.Status.CatalogSourceNamespace}, catalogsource); err != nil {
+					klog.Warning(err)
+					continue
+				}
+			} else {
+				if pm.Status.CatalogSourceNamespace != namespace {
+					continue
+				}
```
Daniel-Fan, line 189: May I ask that is this condition necessary?

Also I think that it would be better if we could include the subsequent logic `currentCatalog := CatalogSource{}` into `if hasCatalogPermission` section after Catalog is obtained.

controllers/operator/manager.go
```diff
@@ -172,11 +173,22 @@ func (m *ODLMOperator) GetCatalogSourceAndChannelFromPackage(ctx context.Context
...
 			}
+
 			catalogsource := &olmv1alpha1.CatalogSource{}
-			if err := m.Reader.Get(ctx, types.NamespacedName{Name: pm.Status.CatalogSource, Namespace: pm.Status.CatalogSourceNamespace}, catalogsource); err != nil {
-				klog.Warning(err)
-				continue
+			hasCatalogPermission := m.CheckResAuth(ctx, namespace, "operators.coreos.com", "catalogsources", "get")
+			if hasCatalogPermission {
+				// Fetch the CatalogSource if cluster permission allows
+				catalogsource := &olmv1alpha1.CatalogSource{}
```
Daniel-Fan, line 181: We had a re-declaration for variable `catalogsource` inside the if condition. The `catalogsource := &olmv1alpha1.CatalogSource{}` in the outside does not have value assigned.

controllers/operator/manager_test.go
```diff
@@ -304,6 +325,7 @@ func TestGetCatalogSourceAndChannelFromPackage(t *testing.T) {
 
 	operator := &ODLMOperator{
 		Reader: fakeReader,
+		Client: fake.NewClientBuilder().Build(),
```
Daniel-Fan, line 328: Can we try something like?
```
fakeReader.On("Create", ctx, mock.Anything, mock.Anything).Return(nil).Run(func(args mock.Arguments) {
		sar := args.Get(1).(*authorizationv1.SubjectAccessReview)
		sar.Status.Allowed = false
	})
```

controllers/operator/manager_test.go
```diff
@@ -400,6 +422,69 @@ func TestGetCatalogSourceAndChannelFromPackage(t *testing.T) {
...
+
+func (m *MockClient) Create(ctx context.Context, obj client.Object, opts ...client.CreateOption) error {
+	args := m.Called(ctx, obj, opts)
+	return args.Error(0)
+}
+
+func TestCheckResAuth(t *testing.T) {
+	ctx := context.TODO()
+
+	mockClient := &MockClient{}
```
Daniel-Fan, line 447: Is it possible to use the above `mockReader := &MockReader{}` to replace `mockClient := &MockClient{}`?

Daniel-Fan (approved):
/lgtm

PR #1103, opened 2025-01-14
add required OLM metadata

pgodowski:
..as defined in https://docs.redhat.com/en/documentation/red_hat_software_certification/2024/html-single/red_hat_openshift_software_certification_policy_guide/index#con-operator-requirements_openshift-sw-cert-policy-products-managed and required by RH Certification


YCShen1010 (approved):
/lgtm

PR #1105, opened 2025-01-14
prototype ODLM w/o olm

bluzarraga:
**What this PR does / why we need it**: to enable install without olm we need odlm to configure operands without using olm. Created a new controller to handle this change. In no olm environments (defined with an environment variable), we use a different controller to reconcile operand requests (called operandrequestnoolm) that looks for a specific annotation (`packageName: /<package name from operandregistry/>) to exist in the namespaces deployments. These deployments are now the home of an operator's alm-examples and odlm then processes these alm-examples the same way it would if it was checking the csv. I have gotten this to deploy and reconcile as expected locally

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/65443

**Release note**:

```release-note

```


bitscuit (approved):
/lgtm

bitscuit:
Ignoring the linter errors for now since code will be refactored to handle later.

PR #1108, opened 2025-01-31
refactor no olm controller code to handle cleanup on operandrequest deletion

bluzarraga:
**What this PR does / why we need it**: This PR enables ODLM to clean up operands and their resources when an operandrequest is deleted in a no olm scenario. I have tried to mirror or reuse as much as the regular olm-enabled process as possible so that it should be working mostly the same. The main difference is that instead of enabling all of this through subscription objects, it is now done through deployment and what I am referring to as a tracking configmap. @bitscuit and I discussed and we do not want to disrupt how ODLM handles knowing what to clean up too much by doing away with the annotation based tracking. However, we do not want these annotations to live in a resource controlled by helm (ie the deployment that is now the center piece of an operator) so we decided to create empty configmaps that are named after the operator's package name and place the same annotations there instead of in the subscription. ODLM should behave the same way in that it will check the CM for the same set of annotations it would have checked the subscription for. It will also update the CM with the same annotations it would have updated the sub with. It may be more appropriate to create a new CRD for a more succinct/specific tracking measure but for now this should work. 

As of this writing (1.31.25) I have not been able to test the code changes yet.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ OPERATOR_VERSION ?= 4.3.9
 # Kind cluster name
 KIND_CLUSTER_NAME ?= "odlm"
 # Operator image tag for test
-OPERATOR_TEST_TAG ?= nolm-controller3
+OPERATOR_TEST_TAG ?= nolm-controller-cleanup
 
 # Options for 'bundle-build'
 ifneq ($(origin CHANNELS), undefined)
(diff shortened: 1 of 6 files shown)


bluzarraga:
I'm not surprised by the linter

bitscuit (approved):
/lgtm

PR #1114, opened 2025-02-04
[WIP] Enhance cleanup of OperandRequest for non-OLM case

YCShen1010:
Test and enhance the cleanup logic for uninstalling operator resources

**What include**:
1. Updated the annotation in the operator ConfigMap by adding or deleting the related OperandRequest. If there are no more requests for that operator, the ConfigMap will be deleted as well.

**Continue Testing**

### DIFF
--- a/helm/templates/rbac.yaml
+++ b/helm/templates/rbac.yaml
@@ -169,6 +169,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: operand-deployment-lifecycle-manager
+  namespace: {{ .Values.operatorNamespace }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: Role
--- a/controllers/operandrequestnoolm/reconcile_operator.go
+++ b/controllers/operandrequestnoolm/reconcile_operator.go
@@ -360,6 +360,12 @@ func (r *Reconciler) uninstallOperatorsAndOperands(ctx context.Context, operandN
 		} else if deploymentList != nil {
 			klog.Infof("Found %d Deployment for package %s/%s", len(deploymentList), op.Name, namespace)
 			if uninstallOperand {
+				// delete the configmap
+				klog.V(1).Infof("Deleting the Configmap %s/%s", cm.Namespace, cm.Name)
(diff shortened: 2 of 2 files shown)


helm/templates/rbac.yaml
```diff
@@ -169,6 +169,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: operand-deployment-lifecycle-manager
+  namespace: {{ .Values.operatorNamespace }}
```
bitscuit, line 172:
```suggestion
  namespace: {{ .Values.global.operatorNamespace }}
```

Actually, the latest code from master branch has already fixed this

controllers/operator/manager.go
```diff
@@ -466,36 +466,73 @@ func (m *ODLMOperator) GetSubscription(ctx context.Context, name, operatorNs, se
...
-			Namespace: ns,
-		}); err != nil {
-			return nil, err
-		}
-
-		for _, dep := range depList.Items {
-			if dep.Name == name || dep.Annotations["packageName"] == packageName {
-				depCandidates = append(depCandidates, dep)
-			}
+	deploy := &appsv1.Deployment{}
```
bitscuit, line 469: Doesn't OLM-based ODLM use the `GetDeployment` function? If so, all of the changes in this function need to be separated out into a function dedicated for non-OLM

bitscuit: I'm assuming those tests that are failing are probably from the changes in this function

helm/templates/rbac.yaml
YCShen1010: Yes, I will do the rebase.

controllers/operator/manager.go
YCShen1010: Right, this `GetDeployment` function is newly created and only used for the non-OLM controller. But since the deployment names are quite different, I will remove it and use `GetDeploymentListFromPackage` instead to find the list of deployments and locate the intended one.

`GetDeploymentListFromPackage` has been used in several places, only in the non-OLM controller. I think it could also used as a re-req check for ODLM to verify if the required deployment is present. If not, it should return an error

bitscuit (approved):
/lgtm

PR #1125, opened 2025-02-14
Create all role and rolebindings in each namespace

YCShen1010:
All charts created the operator's RBAC in all the namespaces for the tenant instead of namespace scope operator doing the permission projection.

**Slack discussion:** https://ibm-cloudplatform.slack.com/archives/C085K4TSJ5S/p1739389039395679

**How to test**:
1. Running `helm template ./helm > tmp.yaml` to check if all the values passed in correctly
2. Apply the helm charts on OCP without NSS operator, and check if ODLM working properly.

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -5,6 +5,7 @@
 global:
   tetheredNamespace: [tenant1,tenant2]
   operatorNamespace: operator-ns
+  instanceNamespace: services-ns
 
 # Note there are no leading or trailing /'s
 cpfs:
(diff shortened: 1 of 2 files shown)


helm/values.yaml
```diff
@@ -1,6 +1,12 @@
 # imagePullPrefix: icr.io
 # imagePullSecret: ibm-entitlement-key
 
+# The following global values could be overridden by top-level values
+global:
+  tetheredNamespace: [tenant1,tenant2]
```
bitscuit, line 6:
```suggestion
  tetheredNamespace: []
```
leave it empty by default

Daniel-Fan (approved):
/lgtm

PR #1149, opened 2025-04-09
Refactor Bindinfo controller logic on object event watching and reconciliation triggering condition

Daniel-Fan:
### Context
When an object like configMap, secret watched by ODLM has an event happens, ODLM will receive this object event and trigger the OperandBindinfo reconciliation related to this Object.

For example, when ConfigMap `ibmcloud-cluster-info` is updated by IAM operator, ODLM would receive such configmap update event, and trigger OprandBindinfo `ibm-iam-bindinfo` reconciliation because ConfigMap `ibmcloud-cluster-info` contains Bindinfo related label `<ns>.ibm-iam-bindinfo/bindinfo: 'true'`.
```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: ibmcloud-cluster-info
  namespace: cd-daily
  labels:
    app: auth-idp
    cd-daily.ibm-iam-bindinfo/bindinfo: 'true'
    operator.ibm.com/managedBy-opbi: original
    operator.ibm.com/watched-by-odlm: 'true'
data:
    ....
```

As an OperandBindinfo reconciliation result, ODLM will update ConfigMap `ibm-iam-bindinfo-ibmcloud-cluster-info` with the same content updated in ConfigMap `ibmcloud-cluster-info`.
```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: ibm-iam-bindinfo-ibmcloud-cluster-info
  namespace: cd-daily
  labels:
    app: auth-idp
    cd-daily.ibm-iam-bindinfo/bindinfo: 'true'
    operator.ibm.com/managedBy-opbi: copy
    operator.ibm.com/watched-by-odlm: 'true'
data:
    .....
```

### Issue
When the ConfigMap `ibmcloud-cluster-info` was updated by IAM operator at the first time, ODLM is not able to trigger OperandBindinfo reconciliation, because the ConfigMap `ibmcloud-cluster-info` object received by ODLM via watch events does contain any labels. Subsequently, ODLM could not identify its corresponding OperandBindinfo `ibm-iam-bindinfo`.

**Which issue(s) this PR fixes:
Fixes
- https://github.ibm.com/PrivateCloud-analytics/CPD-Quality/issues/44061
- https://github.ibm.com/IBMPrivateCloud/roadmap/issues/66378
- https://github.ibm.com/PrivateCloud-analytics/CPD-Quality/issues/44033

### Solution
[...]

PR #1154, opened 2025-04-14
[SC2] Refactor Bindinfo controller logic on object event watching and reconciliation triggering condition

Daniel-Fan:
cherry-pick #1149


YCShen1010 (approved):
/lgtm

PR #1236, opened 2026-04-30
fix Race Condition in Resource Cleanup 

qpdpQ:
During upgrade scenarios (e.g., EDB → IBM PG), when multiple OperandBindInfos reference the same Secret/ConfigMap, a race condition can occur:

In most cases, uninstallation is faster than installation. As a result, the uninstallation removes the existing configmap im-datastore-edb-cm, and the subsequent installation creates a new one, so everything works as expected.

However, in this case, the installation completed faster than the uninstallation. When the installation attempted to create the ConfigMap, it already existed, so no action was taken. Shortly afterward, the uninstallation deleted the ConfigMap. This caused the IM operand pods to fail to find im-datastore-edb-cm and become stuck.

 **Solution**

Implemented cleanup logic in the `cleanupCopies()` function that:

1. Checks for other managers before deleting resources
2. Removes only the current OperandBindInfo's label if other OperandBindInfo exist
3. Deletes the resource only when no other owner remain
4. Cleans up labels on original resources

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69299

### DIFF
--- a/controllers/operandbindinfo/operandbindinfo_controller.go
+++ b/controllers/operandbindinfo/operandbindinfo_controller.go
@@ -739,6 +739,47 @@ func nestedBasicType(obj map[string]interface{}, fields ...string) (string, bool
 	}
 }
 
+// hasOtherManagers checks if there are other OperandBindInfos managing this resource
+// by examining labels with pattern: namespace.name/bindinfo: "true"
+func (r *Reconciler) hasOtherManagers(ctx context.Context, labels map[string]string, ownerRefs []metav1.OwnerReference, currentBindInfoNs, currentBindInfoName string) bool {
+	currentLabel := currentBindInfoNs + "." + currentBindInfoName + "/bindinfo"
+
+	// Check labels for other OperandBindInfos
+	for key, value := range labels {
+		if strings.HasSuffix(key, "/bindinfo") && value == "true" && key != currentLabel {
+			klog.V(3).Infof("Resource has other OperandBindInfo manager: %s", key)
+			return true
+		}
+	}
+
+	return false
+}
+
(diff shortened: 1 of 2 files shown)


qpdpQ:
Test image `quay.io/yuchen_li1/odlm:69299`
Test steps:

1. create two operandreqeusts, one request ibm-pg and another request EDB
Screenshot 2026-04-29 at 8 25 31 PM
Screenshot 2026-04-29 at 8 24 03 PM
Screenshot 2026-04-29 at 8 23 56 PM


2.  Delete one operandrequest, copied secret won't be deleted, but the label and ownerreference will be updated
Screenshot 2026-04-29 at 8 26 01 PM

controllers/operandbindinfo/operandbindinfo_controller.go
```diff
@@ -739,6 +739,47 @@ func nestedBasicType(obj map[string]interface{}, fields ...string) (string, bool
 	}
 }
 
+// hasOtherManagers checks if there are other OperandBindInfos managing this resource
+// by examining labels with pattern: namespace.name/bindinfo: "true"
+func (r *Reconciler) hasOtherManagers(ctx context.Context, labels map[string]string, ownerRefs []metav1.OwnerReference, currentBindInfoNs, currentBindInfoName string) bool {
```
Daniel-Fan: @qpdpQ Can you please simplify this function that there are several arguments not used.

If it is a utility function only, we do not need it to be a method under `r *Reconciler`

Daniel-Fan (changes requested):

Daniel-Fan (approved):
LGTM

PR #1242, opened 2026-05-25
Optimize ODLM controller caching and resource watching

Daniel-Fan:

This PR optimizes the ODLM controller's caching and watching behavior to improve performance and reduce memory consumption. It makes two key changes:

1. **Cache Optimization**: Adds resources to the cache configuration:
   - ClusterServiceVersion resources are cached to support deletion actions
   - Subscription resources are cached to handle both labeled and unlabeled resources

2. **Predicate Filtering**: Implements label-based filtering for Secret and ConfigMap watches in the OperandBindInfo controller. Instead of watching all Secrets and ConfigMaps in watched namespaces, the controller now only watches resources that have the ODLM bindinfo label (`constant.OpbiTypeLabel`). This significantly reduces the number of cached resources and prevents unnecessary reconciliation triggers.

Fixes https://github.ibm.com/PrivateCloud-analytics/CPD-Quality/issues/98796

**Verification**:

Memory usage after applying this fix is within acceptable limits:

```bash
$ oc adm top pod operand-deployment-lifecycle-manager-54bd6f8f6f-d8dvv -n operator-ns
NAME                                                    CPU(cores)   MEMORY(bytes)   
operand-deployment-lifecycle-manager-54bd6f8f6f-d8dvv   78m          115Mi
```

The ODLM pod is running with 115Mi memory usage, which is well below the typical memory limits, confirming that the optimization successfully reduces memory consumption.

### DIFF
--- a/controllers/k8sutil/cache.go
+++ b/controllers/k8sutil/cache.go
@@ -40,9 +40,6 @@ func NewODLMCache(isolatedModeEnable bool, opts ctrl.Options) ctrl.Options {
 	cacheFreshLabelSelector := labels.SelectorFromSet(
 		labels.Set{constant.BindInfoRefreshLabel: "enabled"},
 	)
-	cacheOpreqLabelSelector := labels.SelectorFromSet(
-		labels.Set{constant.OpreqLabel: "true"},
-	)
 
 	watchNamespace := util.GetWatchNamespace()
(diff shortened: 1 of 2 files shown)


qpdpQ:
PR looks good to me! Thank you!

qpdpQ (approved):

PR #1246, opened 2026-06-10
fix: prevent concurrent map access in predicate functions

Daniel-Fan:
What this PR does / why we need it:
This PR fixes race conditions and object mutation issues in the controller code by:
- Preventing concurrent map access in `toOpbiRequest` by immediately capturing labels from the event object instead of initializing as nil and reading later
- Removing unnecessary goroutines in race tests that were modifying shared objects
- Creating defensive copies in `ObjectIsUpdatedWithException` to prevent mutation of input objects
- Introducing a `filteredMetadata` helper function to safely filter metadata maps without modifying the originals
- Adding test coverage to verify objects are not mutated and labels are correctly used even when objects are deleted

Fix: https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69503


Daniel-Fan:
Image `quay.io/daniel_fan/odlm:4.5.9` was tested, indicating there is no pod restart.

YCShen1010 (approved):
/lgtm

PR #1262, opened 2026-08-25
Allow ODLM operator to run without DaemonSet permissions

YCShen1010:
**What this PR does / why we need it**: Enhance the ODLM so that DaemonSet pod refresh is optional and missing DaemonSet permissions do not fail OperandBindInfo reconciliation.

This change allows restricted installations to omit DaemonSet RBAC while preserving the existing behavior by default.

Fixes # https://github.ibm.com/IBMPrivateCloud/roadmap/issues/69869
 
**Changes**:

- When `list/watch/update` permissions are available, keep the existing DaemonSet refresh behavior.
- When permissions are unavailable, log a warning and skip DaemonSet refresh without failing OperandBindInfo reconciliation.
- Continue refreshing Deployments and StatefulSets.
- Add a Helm opt-out value and omit the DaemonSet RBAC rule when the feature is disabled.
- Document that opted-in DaemonSet workloads will require a manual restart after bindinfo Secret/ConfigMap changes when this capability is disabled.


Daniel-Fan (approved):
LGTM

